PR #13, opened 2023-06-23
added pulls option

0ctagon:


0ctagon:
See also https://plothist.readthedocs.io/en/pulls/advanced/advanced_examples.html
Will also add some documentation example with `compare_two_hist` and one with pulls displayed.

plothist/hep_plotters.py
```diff
@@ -17,6 +17,7 @@ def compare_data_mc(
     ylabel=None,
     mc_labels=None,
     mc_colors=None,
+    ratio="diff",
```
cyrraz, line 20: The default is not a difference "diff", but a ratio. In the future we want to implement true difference or residual (data-MC).
We should not call the argument ratio, but `comparison` for example, that could take `ratio`, `pull`, or `diff`. In the future, we may even propose a combination of multiple comparisons with multiple bottom plots, like `["ratio", "pull"]`.

Now come the less fun part: in that direction, we should rename `ax_comparison` -> `ax_main` et `ax_ratio` -> `ax_comparison` for example, unless you have a better idea.

plothist/hep_plotters.py
```diff
@@ -141,21 +145,47 @@ def compare_data_mc(
...
-    scaled_mc_uncertainty = np.sqrt(mc_hist_total.variances()) / mc_hist_total.values()
+    if ratio == "diff":
+        # Compute data/MC ratio
+        ratio_values = np.where(
+            mc_hist_total.values() != 0,
+            data_hist.values() / mc_hist_total.values(),
+            np.nan,
+        )
+        # Compute scaled uncertainties
+        scaled_data_uncertainty = np.sqrt(data_hist.variances()) / data_hist.values()
```
cyrraz, line 156: Looks wrong to me, we should divide by `mc_hist_total.values()` I think

plothist/hep_plotters.py
```diff
@@ -141,21 +145,47 @@ def compare_data_mc(
...
+        scaled_mc_uncertainty = (
+            np.sqrt(mc_hist_total.variances()) / mc_hist_total.values()
+        )
+    elif ratio == "pull":
+        # Compute pulls
+        ratio_values = np.where(
+            data_hist.values() != 0,
+            (mc_hist_total.values() - data_hist.values())
```
cyrraz, line 164: We should have a reference for the definition we use. I think that usually, it is data-mc, not the opposite

plothist/hep_plotters.py
```diff
@@ -141,21 +145,47 @@ def compare_data_mc(
...
+            / np.sqrt(data_hist.variances() + mc_hist_total.variances()),
+            np.nan,
+        )
+        scaled_data_uncertainty = np.where(
+            data_hist.values() != 0,
+            1,
+            np.nan,
+        )
+        scaled_mc_uncertainty = np.where(
```
cyrraz, line 173: I do not think that we should separate data and MC uncertainty in a pull plot. Is it even doable?

plothist/hep_plotters.py
```diff
@@ -165,20 +195,28 @@ def compare_data_mc(
...
         fill=False,
         lw=0,
     )
 
-    ax_ratio.axhline(1, ls="--", lw=1.0, color="black")
-    ax_ratio.set_ylim(0.0, 2.0)
+    if ratio == "diff":
+        ax_ratio.axhline(1, ls="--", lw=1.0, color="black")
+        ax_ratio.set_ylabel(r"$\frac{Data}{Pred.}$")
+        ax_ratio.set_ylim(0.0, 2.0)
```
cyrraz, line 213: We should let the user define the range, this was asked by a German user. If `None` is provided, we set it to (0,2) or (-5,5) depending on `compare`.

plothist/hep_plotters.py
```diff
@@ -165,20 +195,28 @@ def compare_data_mc(
...
+    elif ratio == "pull":
+        ax_ratio.axhline(0, ls="--", lw=1.0, color="black")
+        ax_ratio.set_ylim(-5.0, 5.0)
```
cyrraz, line 216: We should let the user define the range, this was asked by a German user. If `None` is provided, we set it to (0,2) or (-5,5) depending on `compare`.

plothist/plotters.py
```diff
@@ -495,7 +495,14 @@ def plot_hist_difference(hist1, hist2, ax, **kwargs):
 
 
 def compare_two_hist(
```
cyrraz, line 497: Same comments. Shall we try to optimise and have a single function? In that case, hist_1 and hist_2 are not symmetric. One is data and the other one is prediction (or reference). Then we could use call this function in `hep_plotters`

plothist/hep_plotters.py
0ctagon:
```
RuntimeWarning: overflow encountered in double_scalars
  delta = (x1t - x0t) * margin
```

plothist/hep_plotters.py
0ctagon: Need to think about this one

0ctagon:
Should `compare_data_mc()`  be renamed to account the fact that the mc_hist_list argument is a list of mc_contributions (and not a simple mc hist, as we could expect from `compare_data_mc`)?

0ctagon:
`compare_data_mc()` uses `compare_two_hist()`, all the comparison is now computed in `compare_two_hist()` and nothing in `compare_data_mc()`. Needs to check closely the definitions of the uncertainties.

docs/advanced/advanced_examples.rst
```diff
@@ -40,4 +40,16 @@ Publication style:
...
+   :width: 500
+
+
+Color palette
+=============
+
+When displaying quantities such as yields or intensities, it is recommended to utilize Perceptually Uniform Sequential Colormaps like the cubehelix colormap for accurate representation.
+
+We provide the ``cubehelix_palette()`` (see `here <https://plothist.readthedocs.io/en/latest/documentation/documentation.html#plothist.plotters.cubehelix_palette>`_) function to create such colormap. With ``ncolors = 11``:
+
```
cyrraz, line 52: Maybe write the code to produce the colormap plot, including the `from plothist import cubehelix_palette` to help the user.

cyrraz (commented):
Following our bike discussion, we should do this:
* Define `plot_comparison(h1, h2, ax, comparison="ratio")` in `plotters.py`.
* Add the arguments `fig=None`, `ax_main=None`, `ax_comparison=None` to `compare_data_mc()`. If all are None, create the figure and both axes using a new function `create_comparison_figure()`. If none is `None`, use the provided figure and axes. Otherwise, raise an exception.
* Add the arguments `fig=None`, `ax=None`, `plot_mc()`. If all are None, create the figure and the ax. If none is `None`, use the provided figure and ax. Otherwise, raise an exception.
* Call `plot_mc()` and `plot_comparison()` in `plot_data_mc()`.
* We should deprecate `plot_hist_difference()` and add the option `diff` to the argument `comparison` of `plot_comparison`.

cyrraz:
> Should `compare_data_mc()` be renamed to account the fact that the `mc_hist_list` argument is a list of mc_contributions (and not a simple mc hist, as we could expect from `compare_data_mc`)?

The list can contain only one component if needed, so I am not sure we need to change the name

0ctagon:
Every points of the bike discussion have been addressed.

plothist/hep_plotters.py
```diff
@@ -17,115 +22,95 @@ def compare_data_mc(
...
     mc_labels : list of str, optional
-        List of labels for the MC simulations, by default None.
+        The labels for the MC simulations. Default is None.
     mc_colors : list of str, optional
-        List of colors for the MC simulations, by default None.
-    save_as : str, optional
-        File path to save the figure, by default None.
+        The colors for the MC simulations. Default is None.
+    comparison : str, optional
+        The type of comparison to plot ("ratio" or "pull"). Default is "ratio".
```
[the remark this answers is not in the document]
cyrraz: I see now that we force `"split"` treatment. Fine for now. We may change this later if we judge it useful.

plothist/hep_plotters.py
```diff
@@ -137,56 +122,24 @@ def compare_data_mc(
...
-    ax_ratio.set_ylabel(r"$\frac{Data}{Pred.}$")
-
-    _ = ax_comparison.xaxis.set_ticklabels([])
-    fig.subplots_adjust(hspace=0.125)
-
     if save_as is not None:
         fig.savefig(save_as, bbox_inches="tight")
 
-    return fig, ax_comparison, ax_ratio
+    return fig, ax_main, ax_comparison
```
cyrraz, line 142: So clean

plothist/hep_plotters.py
```diff
@@ -197,59 +150,105 @@ def plot_mc(
...
+    fig=None,
+    ax=None,
     save_as=None,
     flatten_2d_hist=False,
+    stacked=True,
+    leg_ncol=1,
 ):
     """
-    Plot Monte Carlo (MC) simulations.
+    Plot MC simulations.
```
cyrraz, line 161: Plot MC simulation histograms.

plothist/hep_plotters.py
```diff
@@ -197,59 +150,105 @@ def plot_mc(
...
+        The Figure object containing the plot.
     ax : matplotlib.axes.Axes
-        Axes instance for the plot.
+        The Axes object containing the plot.
+
     """
 
+    if fig is None and ax is None:
+        fig, ax = plt.subplots()
+    elif (fig is None and ax) or (fig and ax is None):
```
cyrraz, line 203: At this stage, we already know that at least one is not None, so just need to check if at least one is None, and raise the Error in this case

plothist/hep_plotters.py
```diff
@@ -197,59 +150,105 @@ def plot_mc(
...
+            mc_hist_list,
+            ax=ax,
+            color=mc_colors,
+            label=None,
+            stacked=False,
+            histtype="step",
+        )
+        # Plot the sum of the unstacked histograms
+        plot_hist(
+            sum(mc_hist_list),
```
cyrraz, line 246: `mc_hist_total`, as already defined above

plothist/plotters.py
```diff
@@ -9,194 +9,46 @@
...
-
-    with open(path, "r") as f:
-        variable_registry = yaml.safe_load(f)
-        return variable_registry[variable]
-
-
-def update_variable_registry(
-    variable_key, x_min, x_max, path="./variable_registry.yaml"
+def create_comparison_figure(
+    figsize=(6, 4),
```
cyrraz, line 15: `figsize` should be set by the style, not hard coded. None should be the default value, writing in the documentation that it is defined in the style file by default.

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
-        The labels for the two histograms being compared.
-    save_as : str, optional
-        If provided, the filename to save the figure as.
+        The label for the y-axis. Default is None.
+    x1_label : str, optional
+        The label for the first histogram. Default is "x1".
+    x2_label : str, optional
+        The label for the second histogram. Default is "x2".
+    comparison : str, optional
+        The type of comparison to plot. Default is "ratio".
```
cyrraz, line 357: We should tell the user to look at `plot_comparison` for details. Also, `ratio_uncertainty` is missing

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+    save_as : str or None, optional
+        The path to save the figure. Default is None.
+    fig : matplotlib.figure.Figure or None, optional
+        The figure to use for the plot. If fig, ax_main and ax_comparison are None, a new figure will be created. Default is None.
+    ax_main : matplotlib.axes.Axes or None, optional
+        The main axes for the histogram comparison. If fig, ax_main and ax_comparison are None, a new axes will be created. Default is None.
+    ax_comparison : matplotlib.axes.Axes or None, optional
+        The axes for the comparison plot. If fig, ax_main and ax_comparison are None, a new axes will be created. Default is None.
+    **kwargs
+        Additional keyword arguments to be passed to the plot_comparison function.
```
cyrraz, line 369: Interesting, but `kwargs` is not passed to `plot_comparison`

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
     if not np.all(hist_1.axes[0].edges == hist_2.axes[0].edges):
         raise ValueError("The bins of the compared histograms must be equal.")
 
-    fig, (ax_comparison, ax_ratio) = plt.subplots(
-        2, gridspec_kw={"height_ratios": [4, 1]}
-    )
+    if fig is None and ax_main is None and ax_comparison is None:
+        fig, (ax_main, ax_comparison) = create_comparison_figure()
+    elif (fig is not None or ax_main is not None or ax_comparison is not None) and (
+        fig is None or ax_main is None or ax_comparison is None
```
cyrraz, line 391: At this stage, we already know that at least one is not `None`, so just need to check if at least one is `None`, and raise the Error in this case

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
-    ax_comparison.set_xlim(xlim)
-    ax_comparison.set_ylabel(ylabel)
-    ax_comparison.tick_params(axis="x", labelbottom="off")
-    ax_comparison.legend(framealpha=0.5)
+    plot_hist(hist_1, ax=ax_main, label=x1_label, histtype="step")
+    plot_hist(hist_2, ax=ax_main, label=x2_label, histtype="step")
+    ax_main.set_xlim(xlim)
+    ax_main.set_ylabel(ylabel)
+    ax_main.tick_params(axis="x", labelbottom="off")
+    ax_main.legend(framealpha=0.5)
```
cyrraz, line 404: The legend frame alpha should be set by the style, not hard coded

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+    _ = ax_main.xaxis.set_ticklabels([])
+    fig.subplots_adjust(hspace=0.125)
```
cyrraz, line 406: Remove. Already done when creating the figure

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+    ax : matplotlib.axes.Axes
+        The axes to plot the comparison.
+    xlabel : str, optional
+        The label for the x-axis. Default is "x1".
+    x1_label : str, optional
+        The label for the first histogram. Default is "x1".
+    x2_label : str, optional
+        The label for the second histogram. Default is "x2".
+    comparison : str, optional
+        The type of comparison to plot ("ratio" or "pull"). Default is "ratio".
```
cyrraz, line 455: Will make this PR larger, but adding "diff" would be great

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+    comparison_ylim : tuple or None, optional
+        The y-axis limits for the comparison plot. Default is None.
+    ratio_uncertainty : str, optional
+        How to treat the uncertainties of the histograms when comparison = "ratio" ("uncorrelated" for simple comparison, "split" for scaling and split hist_1 and hist_2 uncertainties)
```
cyrraz, line 459: Let's add `"efficiency"` at the next PR.

cyrraz: Let's also say explicitly that this argument has no effect if `comparison` is not `ratio`

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+        ratio_values = np.where(
+            hist_2.values() != 0, hist_1.values() / hist_2.values(), np.nan
+        )
+        if ratio_uncertainty == "split" :
+            ratio_variance = (np.sqrt(hist_1.variances()) / hist_1.values()) ** 2
+        elif ratio_uncertainty == "uncorrelated":
+            ratio_variance = np.where(
+                hist_2.values() != 0,
+                hist_1.variances() / hist_2.values() ** 2
+                + hist_2.variances() * hist_1.values() ** 2 / hist_2.values() ** 4,
```
cyrraz, line 485: We should double check the formula

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
-        hist_2.variances() / hist_1.values() ** 2
-        + hist_1.variances() * hist_2.values() ** 2 / hist_1.values() ** 4,
-        np.nan,
-    )
+    if comparison == "ratio":
+        ratio_values = np.where(
+            hist_2.values() != 0, hist_1.values() / hist_2.values(), np.nan
+        )
+        if ratio_uncertainty == "split" :
+            ratio_variance = (np.sqrt(hist_1.variances()) / hist_1.values()) ** 2
```
cyrraz, line 480: Wrong. This is `h1_scaled_variance`, and we should divide by `hist_2.values()`, this is a bug in the current version of plothist. I understand that calling it `ratio_variance` helps later, but this is wrong name. The ratio variance is computed only in the `uncorrelated` case (and later in the `efficiency` case). We could define `h1_scaled_uncertainty` instead (not taking the square).

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+
     np.seterr(divide="warn", invalid="warn")
 
-    ax_ratio.errorbar(
-        x=hist_1.axes[0].centers,
+    ax.errorbar(
+        x=hist_2.axes[0].centers,
         xerr=None,
-        y=np.nan_to_num(ratio, nan=0),
+        y=np.nan_to_num(ratio_values, nan=0),
```
cyrraz, line 515: Rename to `comparison_values`. Less confusing, because this is also used in the `"pulls"` instead of `"ratio"` argument.

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+            ratio_variance = np.where(
+                hist_2.values() != 0,
+                hist_1.variances() / hist_2.values() ** 2
+                + hist_2.variances() * hist_1.values() ** 2 / hist_2.values() ** 4,
+                np.nan,
+            )
+        else:
+            raise ValueError("ratio_uncertainty not in ['uncorrelated', 'split'].")
+        if ratio_uncertainty == "split" :
+            h2_uncertainty = np.sqrt(hist_2.variances()) / hist_2.values()
```
cyrraz, line 491: Wrong, we should call it `h2_scaled_uncertainty`

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
-    ax_ratio.axhline(1, ls="--", lw=1.0, color="black")
-    ax_ratio.set_ylim(0.0, 1.5)
-    ax_ratio.set_xlim(xlim)
-    ax_ratio.set_xlabel(xlabel)
-    ax_ratio.set_ylabel(r"$\frac{" + x2_label + "}{" + x1_label + "}$")
+    if comparison == "ratio":
+        if ratio_uncertainty == "split":
+            ax.bar(
+                x=hist_2.axes[0].centers,
+                bottom=np.nan_to_num(1 - h2_uncertainty, nan=0),
```
cyrraz, line 525: `h2_scaled_uncertainty`

plothist/plotters.py
```diff
@@ -469,107 +321,237 @@ def plot_error_hist(hist, ax, **kwargs):
...
+                height=np.nan_to_num(2 * h2_uncertainty, nan=100),
```
cyrraz, line 526: `h2_scaled_uncertainty`

plothist/plotters.py
cyrraz: Given the amount of comments below, let's keep it for the next MR.

cyrraz (commented):
Very nice overall, I found a few bugs.

plothist/hep_plotters.py
cyrraz: I do not understand the cause of this warning

0ctagon: Dividing by `np.nan` or `0` caused this.
Added a `np.where()` to get rid of this problem.

plothist/plotters.py
0ctagon: Already in the "See Also" section

plothist/plotters.py
0ctagon: We should find a source

plothist/plotters.py
```diff
@@ -42,6 +42,9 @@ def create_comparison_figure(
         Array of Axes objects representing the subplots.
 
     """
+    if figsize is None:
+        figsize = plt.rcParams["figure.figsize"]
```
cyrraz, line 46: Not needed, the rcParams are read by `plt.subplots()` internally

plothist/plotters.py
```diff
@@ -42,6 +42,9 @@ def create_comparison_figure(
...
+
     fig, axes = plt.subplots(
         nrows=nrows, ncols=ncols, figsize=figsize, gridspec_kw=gridspec_kw
```
cyrraz, line 49: remove `figsize=figsize`

plothist/plotters.py
```diff
@@ -512,18 +516,26 @@ def plot_comparison(
...
+        if comparison_ylim is None:
+            comparison_ylim = (0.0, 2.0)
+        ax.set_ylim(comparison_ylim)
+
         if ratio_uncertainty == "split":
             ax.bar(
                 x=hist_2.axes[0].centers,
-                bottom=np.nan_to_num(1 - h2_uncertainty, nan=0),
-                height=np.nan_to_num(2 * h2_uncertainty, nan=100),
+                bottom=np.nan_to_num(1 - h2_scaled_uncertainty, nan=0),
```
cyrraz, line 535: More general: `nan=comparison_ylim[0]` (in case of negative ratio)

plothist/plotters.py
```diff
@@ -512,18 +516,26 @@ def plot_comparison(
...
+        if (ratio_uncertainty == "uncorrelated" or comparison == "pull")
+        else np.nan_to_num(h1_scaled_uncertainty, nan=0),
         fmt=".",
         color="black",
     )
 
     if comparison == "ratio":
+        if comparison_ylim is None:
+            comparison_ylim = (0.0, 2.0)
+        ax.set_ylim(comparison_ylim)
```
cyrraz, line 530: This should be done only once at the end of the function, together with `set_xlim()`

plothist/plotters.py
```diff
@@ -532,20 +544,15 @@ def plot_comparison(
...
         ax.axhline(0, ls="--", lw=1.0, color="black")
         ax.set_ylabel(
             rf"$\frac{{ {x1_label} - {x2_label} }}{{ \sqrt{{\sigma^2_{{{x1_label}}} + \sigma^2_{{{x2_label}}}}} }} $"
         )
         if comparison_ylim is None:
-            ax.set_ylim(-5.0, 5.0)
-        else:
-            ax.set_ylim(comparison_ylim)
+            comparison_ylim = (-5.0, 5.0)
+        ax.set_ylim(comparison_ylim)
```
cyrraz, line 555: This should be done only once at the end of the function, together with `set_xlim()`

plothist/plotters.py
```diff
@@ -512,18 +516,26 @@ def plot_comparison(
     ax.errorbar(
         x=hist_2.axes[0].centers,
         xerr=None,
-        y=np.nan_to_num(ratio_values, nan=0),
-        yerr=np.nan_to_num(np.sqrt(ratio_variance), nan=0),
+        y=np.nan_to_num(comparison_values, nan=0),
```
cyrraz, line 519: I was hoping that keeping nans would work (just not plot these points)

plothist/plotters.py
```diff
@@ -512,18 +516,26 @@ def plot_comparison(
...
+        yerr=np.nan_to_num(np.sqrt(ratio_variance), nan=0)
```
cyrraz, line 520: I was hoping that keeping nans would work (just not plot these uncertainties)

plothist/plotters.py
```diff
@@ -512,18 +516,26 @@ def plot_comparison(
...
+        if (ratio_uncertainty == "uncorrelated" or comparison == "pull")
+        else np.nan_to_num(h1_scaled_uncertainty, nan=0),
```
cyrraz, line 522: I was hoping that keeping nans would work (just not plot these uncertainties)

cyrraz (commented):
Nearly there, just a few last comments. Could you please also update (find+replace) the documentation since we changed our convention for axis names:
* `ax_comparison` -> `ax_main`
* `ax_ratio` -> `ax_comparison`

cyrraz (approved):

PR #14, opened 2023-06-29
added simple examples

0ctagon:


0ctagon:
See https://plothist.readthedocs.io/en/docs/basics/1d_hist.html#adding-simple-elements

cyrraz (approved):

PR #15, opened 2023-07-10
Add documentation to update plothist

cyrraz:
Add a few instructions to update plothist. Will be useful since our users want to use the pulls.

### DIFF
--- a/docs/usage/installation.rst
+++ b/docs/usage/installation.rst
@@ -1,8 +1,8 @@
 .. _installation-label:
 
-============
-Installation
-============
+=======================
+Installation and update
+=======================
 
 
 Using PyPI (stable version)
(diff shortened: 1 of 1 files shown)


docs/usage/installation.rst
```diff
@@ -37,6 +37,7 @@ Using PyPI (stable version)
 
    The ``plothist`` package can now be imported in any of your scripts. Now you can go on with the :ref:`basics-1d_hist-label`.
 
+To update ``plothist`` to its latest stable version, follow the instructions above, adding the argument ``--upgrade`` right after ``pip3 install``.
```
0ctagon, line 40: Maybe add a `.. code-block:: python` so the user can just copy/paste?

cyrraz: done

0ctagon (approved):

PR #23, opened 2023-07-19
small style fix and use Gaussian uncertainties if data is un-weighted

cyrraz:


cyrraz:
Let me merge these hotfixes, because they are rather urgent.

PR #28, opened 2023-07-20
Update default_style.mplstyle

cyrraz:
Latin Modern as default font


0ctagon:
Maybe add the scripts to dl the optimized fonts if the user wants to?

PR #29, opened 2023-07-20
Cyrraz patch 1

cyrraz:
New fonts for style


docs/usage/style.rst
```diff
@@ -1,14 +1,16 @@
 .. _advanced-advanced_examples-label:
```
0ctagon, line 1: .. _advanced-style-label:

0ctagon (approved):

PR #34, opened 2023-07-21
comparison difference

0ctagon:


plothist/hep_plotters.py
```diff
@@ -160,6 +160,23 @@ def compare_data_mc(
...
             )
+    elif comparison_kwargs["comparison"] == "difference":
+        data_hist_high = data_hist.copy()
+        data_hist_high[:] = np.stack(
+            [data_hist_high.values(), uncertainties_high ** 2], axis=-1
+        )
+        data_hist_low = data_hist.copy()
+        data_hist_low[:] = np.stack(
+            [data_hist_low.values(), uncertainties_low ** 2], axis=-1
+        )
```
cyrraz, line 171: The 8 lines above are copied and pasted from just above, we should try to avoid this.

plothist/plotters.py
```diff
@@ -485,11 +485,11 @@ def plot_comparison(
     h2_label : str, optional
         The label for the second histogram. Default is "h2".
     comparison : str, optional
-        The type of comparison to plot ("ratio" or "pull"). Default is "ratio".
+        The type of comparison to plot ('ratio', 'pull' or 'difference'). Default is "ratio".
```
cyrraz, line 488: double quotation marks

plothist/plotters.py
```diff
@@ -537,9 +537,12 @@ def plot_comparison(
             np.nan,
         )
         comparison_variances = np.ones_like(comparison_values)
+    elif comparison == "difference":
+        comparison_values = hist_1.values() - hist_2.values()
+        comparison_variances = np.sqrt(hist_1.variances() + hist_2.variances())
```
cyrraz, line 542: Wrong! The variance is not the sqrt of the sum, but only the sum

cyrraz (commented):
Very nice PR, could you please add an example in the documentation. By doing so, we also have a test of the new option.

plothist/hep_plotters.py
0ctagon: I think it's fine for now, but we can think of a optimized way for future version.

cyrraz (approved):

PR #45, opened 2023-09-19
Compare data fit

miverone:
Here's some additions to plot fit and asymmetries


cyrraz:
Hello @miverone, thank you very much for this PR and sorry for the time it took to close the issue it was raising. The code of plothist was largely modified since the opening of this PR and we could not take the code of this PR as is, since the code logic is now quite different. Good news is that version 1.0 of plothist supports models made of histograms or functions, and also asymmetry comparison.

PR #51, opened 2023-10-02
added matplotlib simple example

0ctagon:


0ctagon:
https://plothist.readthedocs.io/en/newexample/usage/style.html

cyrraz (commented):
Good idea, but "Entries" should be used instead of "Candidates". I do not know what a "candidate" is, even I guess it is not a political election since there are >100_000 "candidates".

cyrraz (approved):

PR #55, opened 2023-10-10
add text function added

0ctagon:


plothist/hep_plotters.py
```diff
@@ -427,9 +427,9 @@ def add_luminosity(
     collaboration : str, optional
         Collaboration name, by default "Belle II"
     x : float, optional
-        x position, by default 1.0.
+        x position, by default "right" = 1.0.
```
cyrraz, line 430: x position in unit of the x-axis length. The default is value "right", which is an alias for 1.0.

plothist/plothist_style.py
```diff
@@ -201,3 +201,72 @@ def get_fitting_ylabel_fontsize(ax):
...
+    Add text to an axis.
+
+    Parameters
+    ----------
+    text : str
+        The text to add.
+    x : float, optional
+        x position, can be a float in [0, 1] or "left"/"right" to put the text to the left (x=0) or right (x=1) of the plot, by default "left".
+    y : float, optional
+        y position, can be a float in [0, 1] or "top"/"bottom" to put the text to the top (y=1.01) or bottom (y=0) of the plot, by default "top".
```
cyrraz, line 225: It is not necessarily in [0,1]. What about

y position in unit of the y-axis length. The default is value "top", which is an alias for 1.01. The other alias "bottom" corresponds to 0.0.

cyrraz (changes requested):

cyrraz (approved):

PR #57, opened 2023-10-19
Savefig function, better 2D hist and utility documentation

0ctagon:
I open this pull request to have a base for discussion about a function to save plots with correct figsize, subplots positions and dimensions.
The function provided here works wonderfully for any 1d or comparisons plots. It doesn't work at all for 2D plots (the colorbar is always out).
We will learn a lot about matplotlib limits, whether this PR will be merged or not.


0ctagon:
After further investigation, it doesn't work with 2D plots because the figsize is too small. We specify a squared figsize to get a square 2D hist, but the final figsize has to be larger to at least display the label of the colorbar.

0ctagon:
https://plothist.readthedocs.io/en/savefig/usage/utilities.html

cyrraz:
Let's discuss it, I would not merge this function if it is not compatible with all the currently available figures.

0ctagon:
Now, everything works with the custom `savefig`. For 2D plots, `adjust_after_tight_layout` has to be used to adjust the space between `ax_colorbar` and `ax`.

0ctagon:
https://plothist.readthedocs.io/en/savefig/usage/utilities.html#save-figure gif added :smile:

plothist/plothist_style.py
```diff
@@ -237,25 +240,25 @@ def add_text(
...
     transform = ax.transAxes
 
-    if x=="left":
+    if x == "left":
         x = 0.0
-    elif x=="right":
+    elif x == "right":
         x = 1.0
-    elif type(x)!=float or type(x)!=int:
+    elif type(x) != float or type(x) != int:
```
cyrraz, line 254: `type(x) not in [float, int]`

plothist/plothist_style.py
```diff
@@ -237,25 +240,25 @@ def add_text(
...
         raise ValueError(f"x should be a float or 'left'/'right' ({x} given))")
 
-    if y=="top":
+    if y == "top":
         y = 1.01
-    elif y=="bottom":
+    elif y == "bottom":
         y = 0.0
-    elif type(y)!=float or type(y)!=int:
+    elif type(y) != float or type(y) != int:
```
cyrraz, line 261: `type(y) not in [float, int]`

plothist/plotters.py
```diff
@@ -281,30 +282,53 @@ def plot_hist(hist, ax, **kwargs):
...
-        divider = make_axes_locatable(ax)
-        ax_colorbar = divider.append_axes("right", size="5%", pad=0.05)
     pcolormesh_kwargs.setdefault("edgecolors", "face")
+
+    if fig is None and ax is None and ax_colorbar is None:
+        fig, (ax, ax_colorbar) = plt.subplots(
+            figsize=(6, 4.5), ncols=2, gridspec_kw={"width_ratios": [4, 0.23]}
+        )
+    elif fig is None or ax is None or ax_colorbar is None:
+        raise ValueError("Need to provid fig, ax and ax_colorbar (or None of them).")
```
cyrraz, line 321: None -> none

plothist/plotters.py
```diff
@@ -758,6 +781,61 @@ def plot_comparison(
...
+    for k_ax, ax in enumerate(axes):
+        ax.set_position(
+            Bbox.from_bounds(
+                ax.get_position().x0,
+                ax.get_position().y0,
+                axes_dimensions[k_ax][0],
+                axes_dimensions[k_ax][1],
+            )
+        )
+    if adjust_after_tight_layout:
```
cyrraz, line 831: why not always doing it?

cyrraz (changes requested):
Just a few comments, but nearly ready to be merged

plothist/plotters.py
0ctagon: Always doing it can mess up some plots. Not doing it cannot mess up plots, only make some subplots misaligned on a very few examples.

cyrraz (approved):

PR #58, opened 2023-10-23
Fix error computation bugs

0ctagon:
Fixed a bug when using `relative_difference` as comparison in `compare_data_mc`. Solution is rather simple.
Uncertainties are the same as `ratio` for both `ratio_uncertainty` options.


plothist/hep_plotters.py
```diff
@@ -145,7 +145,7 @@ def compare_data_mc(
             uncertainties_high**2,
         )
         data_hist[:] = np.stack([data_hist.values(), data_variances], axis=-1)
-    elif comparison_kwargs["comparison"] == "ratio":
+    elif comparison_kwargs["comparison"] in ["ratio", "relative_difference"]:
```
cyrraz, line 148: After all the `elif`, we should add an `else` that raises a `NotImplementedError` to avoid the same kind of bugs in the future if we add new comparisons

cyrraz (changes requested):

0ctagon:
In the last commit `pull comparison_data_MC bugfix`, I correct a bug appearing when using pull as a comparison method in `compare_data_mc`.

Before, when we used `pull` as a comparison method in `compare_data_mc`, we modify the uncertainties of `data_hist`:

```
if comparison_kwargs["comparison"] == "pull":
    data_variances = np.where(
        data_hist.values() >= mc_hist_total.values(),
        uncertainties_low**2,
        uncertainties_high**2,
    )
    data_hist[:] = np.stack([data_hist.values(), data_variances], axis=-1)
```

This, when using `data_hist` after in other comparisons, will modify the result of the comparison.

See example attached, when is plotted `ratio`, `pull` and again `ratio` comparisons before and after the fix. `ratio` is used to show the bug, but it affects every comparison that use `data_hist`.

I think the fix still need some work. When we `plot_error_hist` the `data_hist`, the errors are not the same when using `pull` compare to using any other method.

0ctagon:
https://plothist.readthedocs.io/en/relative_difference_datamc_fix/advanced/hep_examples.html#other-comparisons

plothist/plothist_style.py
```diff
@@ -248,14 +248,14 @@ def add_text(
...
-    elif type(x)!=float or type(x)!=int:
+    elif type(x)!=float and type(x)!=int:
         raise ValueError(f"x should be a float or 'left'/'right' ({x} given))")
 
     if y=="top":
         y = 1.01
     elif y=="bottom":
         y = 0.0
-    elif type(y)!=float or type(y)!=int:
+    elif type(y)!=float and type(y)!=int:
```
cyrraz, line 258: Super peaky, but more elegant would be `elif type(y) not in [float, int]`

plothist/plothist_style.py
```diff
@@ -248,14 +248,14 @@ def add_text(
         x = 0.0
     elif x=="right":
         x = 1.0
-    elif type(x)!=float or type(x)!=int:
+    elif type(x)!=float and type(x)!=int:
```
cyrraz, line 251: Super peaky, but more elegant would be `elif type(y) not in [float, int]`

docs/advanced/hep_examples.rst
```diff
@@ -171,7 +171,68 @@ If you do not want to show and take into account the MC uncertainties, setting `
...
+    fig, axes = create_comparison_figure(
+        figsize=(6, 11),
+        nrows=5,
+        gridspec_kw={"height_ratios": [3.3, 1, 1, 1, 1]},
+        hspace=0.3,
+    )
+
+    fig_temp, ax_temp = plt.subplots()
+
+    for k_comp, comparison in enumerate(["ratio", "pull", "relative_difference", "difference"], 1):
```
cyrraz, line 199: since this figure helped us a lot to catch a bug, I would add two other comparisons: ratio and relative difference with uncorrelated uncertainty instead of split.

cyrraz (changes requested):
Very nice PR

cyrraz (approved):
Hello @0ctagon, I accept this PR. I left two minor comments, but you could merge without implementing them in case you decide so.

PR #61, opened 2023-10-23
remove link to un-existing script

0ctagon:
Title.
We don't have the script anymore. If the command line doesn't work, the other alternative should be to download manually the fonts.

### DIFF
--- a/docs/usage/style.rst
+++ b/docs/usage/style.rst
@@ -28,16 +28,11 @@ Fonts
 
 Latin Modern fonts are used by default (Latin Modern Math, Latin Modern Roman, Latin Modern Sans).
 
-You can install the fonts by running on your terminal:
-
-On Linux/Ubuntu/MacOS
----------------------
+On Linux/Ubuntu/MacOS, you can install the fonts by running on your terminal:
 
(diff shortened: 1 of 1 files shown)


docs/usage/style.rst
```diff
@@ -28,17 +28,12 @@ Fonts
...
-
-On Linux/Ubuntu/MacOS
----------------------
+On Linux/Ubuntu/MacOS, you can install the fonts by running on your terminal:
 
 .. code-block:: bash
 
     install_latin_modern_fonts
 
-If the command above does not work, you can install the fonts by using the script located in the `github repository of plothist <https://github.com/cyrraz/plothist/tree/main/scripts/install_latin_modern_fonts.py>`_.
```
cyrraz, line 40: We should update the link, not remove it:

`https://github.com/cyrraz/plothist/blob/main/plothist/scripts/install_latin_modern_fonts.py`

The idea is that if the command does not work, the user can read the details of the script that install the fonts

cyrraz (changes requested):

cyrraz (approved):

PR #63, opened 2023-11-03
update figures in doc

cyrraz:
In two svg, the date and version of matplotlib appears. @0ctagon could you please edit the figure generator?

Check the difference of the svg sources for details.


0ctagon:
The automatic-plot-script was updated to not have this problem in the futur.

0ctagon (approved):

PR #66, opened 2023-11-17
Option to have the right edge inclusive

0ctagon:
Proposition to add the option of including the right edge of the axis.

This behavior was implemented in `boost-histogram 1.4.0`: https://github.com/scikit-hep/boost-histogram/blob/c08b7efc34b4d9b07c7b6a71bd566fcf303c1d0f/docs/changelog.md?plain=1#L14

Useful when doing discrete variable representation, like a distribution following a bool 0 or 1. The trick we use currently is to set the upper edge to 1.001. With this PR, the upper edge can be 1, with the argument `overflow=False`.

This is not redundant with the implementation of a `axis.__Category`. `boost_histogram` only allows for `int` or `str` categories.

### DIFF
--- a/plothist/plotters.py
+++ b/plothist/plotters.py
@@ -54,7 +54,7 @@ def create_comparison_figure(
     return fig, axes
 
 
-def create_axis(data, bins, range=None):
+def create_axis(data, bins, range=None, overflow=True):
     """
     Create an axis object for histogram binning based on the input data and parameters.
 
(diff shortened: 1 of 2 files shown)


cyrraz:
I see, I tried the script below with boost-histogram-1.3.2 and boost-histogram-1.4.0 and the behaviour is different indeed.

```
import boost_histogram as bh
storage = bh.storage.Weight()
axis = bh.axis.Regular(2, 0, 1, overflow=False)
h = bh.Histogram(axis, storage=storage)
h.axes[0].edges
h.fill(1)
h.view()
```

PR #67, opened 2023-11-17
figure updates with latest matplotlib

cyrraz:
Figures of the main produced with latest matplotlib version.


0ctagon (approved):
Tested on my side, same results.

PR #70, opened 2023-11-21
Plot model

cyrraz:
Define `plot_model()` and `compare_data_model()` as more general versions of `plot_mc()` and `compare_data_mc()`.

The changes on the svg look like a simple ordering in the svg sources.


0ctagon:
Could you show a plot with stacked and unstacked hist on the same?

cyrraz:
> Could you show a plot with stacked and unstacked hist?

Added in the doc

PR #75, opened 2023-11-28
Image gallery

0ctagon:
Adding image gallery in the documentation.
Currently, only `1d_hist.rst` has been implemented.
Waiting for feedback to continue.

See the gallery here: https://plothist.readthedocs.io/en/gallery-test/


cyrraz:
> Adding image gallery in the documentation. Currently, only `1d_hist.rst` has been implemented. Waiting for feedback to continue.
> 
> See the gallery here: https://plothist.readthedocs.io/en/gallery-test/basic_examples_gallery/index.html

Very nice! We should maybe have all the plots on a single page (sorted from less complex to more complex), I think it would generate the best user experience. Also, the gallery could replace most of the sections of the current documentation.

Shall we use the opportunity of this PR to add the dataset generated by `generate_dummy_data()` in the repository (estimated size=8.4 MB) and add a function to get it, so that we can drop the scikit-learn dependence. We could even drop the `pandas` dependence and load the data as numpy arrays, but this we can do later.

0ctagon:
Before putting too much work on a doc that might change a lot with all current PR of v1.0, I will wait until they are merged to continue working on this.

PR #84, opened 2023-12-04
Svg diff after gallery

0ctagon:
Branched from `gallery-test`.

Using the new `make_example` that runs the python scripts directly, some discrepancy is observed on the svg metadata.

Created a PR just to see the differences.


0ctagon:
After checking, the discrepancy comes from incorrect metadata and incorrect ranges from the old script.

PR #85, opened 2023-12-04
remove hard-coded values

cyrraz:
remove hard-coded values in stacked and unstacked kwargs, remove legend_ncols agument

### DIFF
--- a/plothist/plotters.py
+++ b/plothist/plotters.py
@@ -522,7 +522,7 @@ def plot_comparison(
 
     if np.allclose(lower_uncertainties, upper_uncertainties, equal_nan=True):
         hist_comparison = bh.Histogram(hist_2.axes[0], storage=bh.storage.Weight())
-        hist_comparison[:] = np.c_[comparison_values, lower_uncertainties ** 2]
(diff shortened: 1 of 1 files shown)


plothist/plotters.py
```diff
@@ -522,7 +522,7 @@ def plot_comparison(
 
     if np.allclose(lower_uncertainties, upper_uncertainties, equal_nan=True):
         hist_comparison = bh.Histogram(hist_2.axes[0], storage=bh.storage.Weight())
-        hist_comparison[:] = np.c_[comparison_values, lower_uncertainties ** 2]
+        hist_comparison[:] = np.c_[comparison_values, lower_uncertainties**2]
```
0ctagon, line 525: We should get the same `black` version.
I had `21.10b0` and upgraded to the latest `23.11.0`.

0ctagon (approved):

plothist/plotters.py
cyrraz: I updated to 23.10b0, the file stays the same as in the PR

PR #86, opened 2023-12-04
Fix legend order when using ax.stackplot()

cyrraz:
Fix legend order when using ax.stackplot()


0ctagon (commented):
Tested the new `plot_function` in a lot of different examples, the legend is correct every time.

However, when `plot_data_model_comparison` is used and then a simple `plot_function` with stacked, the unstack and the model from  `plot_data_model_comparison` go first in the legend, following by the correct stacked order of both functions.

If this behavior is not problematic, I will approve.

plothist/plotters.py
```diff
@@ -522,7 +552,7 @@ def plot_comparison(
 
     if np.allclose(lower_uncertainties, upper_uncertainties, equal_nan=True):
         hist_comparison = bh.Histogram(hist_2.axes[0], storage=bh.storage.Weight())
-        hist_comparison[:] = np.c_[comparison_values, lower_uncertainties ** 2]
+        hist_comparison[:] = np.c_[comparison_values, lower_uncertainties**2]
```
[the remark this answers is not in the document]
0ctagon: Please upgrade to `23.11.0`.

cyrraz:
> Tested the new `plot_function` in a lot of different examples, the legend is correct every time.
> 
> However, when `plot_data_model_comparison` is used and then a simple `plot_function` with stacked, the unstack and the model from `plot_data_model_comparison` go first in the legend, following by the correct stacked order of both functions.
> 
> If this behavior is not problematic, I will approve.

I think it is corrected, I have also pushed a utility function to reorder the legend entries.

plothist/plotters.py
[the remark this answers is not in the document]
cyrraz: Yes, done

plothist/plothist_style.py
```diff
@@ -353,3 +353,58 @@ def add_luminosity(
...
+    >>> plot_reordered_legend(ax, [1, 0])
+
+    """
+
+    # Extract the original handlers and labels
+    handlers, labels = ax.get_legend_handles_labels()
+
+    # Check if order is valid
+    if not all(i in range(len(labels)) for i in order):
+        raise ValueError(
```
0ctagon, line 400: Need to also check if every number is unique. No error is raised if user input 2x the same number.

docs/usage/utilities.rst
```diff
@@ -40,7 +40,10 @@ Automatic ylabel fontsize
    :alt: Pull comparison
    :width: 500
 
+Reorder legend entries
+======================
 
+The function ``reorder_legend(ax, order)`` reorder the legend entries of a given plot. It is useful when you want to have a specific order in your legend.
```
0ctagon, line 46: Copy/paste the small code example from the doc.

docs/usage/utilities.rst
```diff
@@ -40,6 +40,23 @@ Automatic ylabel fontsize
...
+   import matplotlib.pyplot as plt
+   from plothist import plot_reordered_legend
+
+   fig, ax = plt.subplots()
+   ax.plot([1, 2, 3], label='Line 1')
+   ax.plot([3, 2, 1], label='Line 2')
+
+   # To reorder the legend so that 'Line 2' comes first, use:
+   plot_reordered_legend(ax, [1, 0])
+   plt.show()
```
0ctagon, line 59: If you don't show the plot, remove this line.

0ctagon (approved):

PR #89, opened 2023-12-08
Added doc for roofit to plothist

0ctagon:
This is a PR to add in the documentation how to convert RooFit and zfit pdfs to object that can be used by plothist.
The two functions involved have been tested and reproduce this plot, compared to the Roofit plot:
Screenshot from 2023-12-08 15-59-20
I don't know if we should also put the example to do the plot, with or without the code, in the doc or not. Maybe as a example in the gallery without link in the actual doc except a reference when we describe the 2 functions needed to be used.
The first function needs Roofit/zfit, so it won't be in plothist. The second one is really general, but I don't think it's the role of plothist to have such function.

Link to doc: https://plothist.readthedocs.io/en/add-fit-to-func-doc/usage/misc.html#plot-fit-results


docs/usage/misc.rst
```diff
@@ -0,0 +1,168 @@
...
+      pdf : callable
+         The renormalized PDF.
+      """
+      xmin, xmax = x_range
+      bin_width = (xmax - xmin) / bins
+      integral = quad(pdf, xmin, xmax)[0]
+
+      return lambda x: pdf(x) * n_data * integral * bin_width
+
+Then you can use ``plot_model()`` or ``plot_data_model_comparison()`` to plot the PDF and do all sort of comparisons with the ``plothist`` interface.
```
cyrraz, line 168: * add a link to the documentation explaining how to plot a pull plot with plothist and model made of functions, maybe showing an example of plot here for illustration (but no RooFit plot in my opinion)
* add a link from the documentation explaining how to do a pull plot to this page.

docs/usage/misc.rst
```diff
@@ -0,0 +1,168 @@
...
+      n_data : int
+         The number of data events expected in the PDF range.
+
+      Returns
+      -------
+      pdf : callable
+         The renormalized PDF.
+      """
+      xmin, xmax = x_range
+      bin_width = (xmax - xmin) / bins
```
cyrraz, line 163: This works only for constant bin size.

docs/usage/misc.rst
```diff
@@ -0,0 +1,168 @@
...
+      integral = quad(pdf, xmin, xmax)[0]
```
cyrraz, line 164: Do we not say in the documentation of the function that this number is one?

docs/usage/misc.rst
```diff
@@ -0,0 +1,168 @@
...
+Renormalize the PDF
+-------------------
+
+The ``pdf_func`` you get, by either get it from function or read the saved pickle file for ``RooFit`` or ``zfit``, has an area of 1. When you want to plot it, you need to multiply it by the bin width of your histogram, the number of expected events in the range and the integral of the PDF in the range. This can be done easily using this small function:
+
+.. code-block:: python
+
+   from scipy.integrate import quad
+
+   def renormalize(pdf, x_range, bins, n_data):
```
cyrraz, line 142: Do we really need this two-step method (saving the pdf, then renormalising it)? Can't we do both in one function? Is it because we want the user to have two environments? One for fitting and the other one for plotting?

docs/usage/misc.rst
```diff
@@ -0,0 +1,168 @@
...
+      for k, i in enumerate(x, 1):
+         var.setVal(i)
+         # Evaluate the PDF at the given x value
+         pdf_x.append(pdf.getVal(var))
+         if verbose:
+               print(f"\t{path.replace('.pkl','')} progress [{k*100/len(x):.1f}%]", end="\r")
+      if verbose: print()
+
+      # Interpolate the PDF
+      pdf_func = interp1d(x, pdf_x)
```
cyrraz, line 70: Can we avoid fitting and save the real function?

cyrraz (changes requested):

docs/usage/misc.rst
0ctagon: It is, and it returns 1 if the range was the fitting range. If the user wants to only plot in a subrange of the whole range, he can also input the expected `n_data` in this new range and the renomalization will still work.

docs/usage/misc.rst
0ctagon: I don't think there are easier ways to do that. We want to save a scipy object to input in `plothist`, not a RooFit object.

docs/usage/misc.rst
0ctagon: Renormalization method is the same for `RooFit` and `zfit` once it's transformed to a `scipy` object. Usually, you want to save your pdfs with the integral equal to 1, and then renormalize only when plotting. The user can still do both in 1 function, but in the doc it's better to show this good practice. 
I agree with your last comment, but I finally don't think we need this function as an utility in `plothist`. In the doc it's enough.

cyrraz: I don't think we necessarily want to scale to the number of data entries. Only the pdf of the total model should be scaled that way. So I think the function should take only 2 arguments: `pdf`, `scale_factor`. This also resolves my other comment, because in case of uneven binning, we let the user do the computation. I agree that this can stay outside of plothist. We can show how to compute the scaling factor outside of the function.

While I was writting this comment, I may start understanding better what we are doing: n_data is not the total number of data events in that range but the number of predicted data events in that range corresponding to a certain model component. If this is the case, the function can stay as is, but I would add this explanation in the documentation of the `n_data` argument and I would also mention in the documentation that a regular binning is assumed. In that regard, we should rename `bins` -> `n_bins`.

docs/usage/misc.rst
cyrraz: See my new comment.

docs/advanced/other_advanced.rst
```diff
@@ -30,10 +30,11 @@ First, we make the histograms and scale them. Then, we plot the histograms and t
    :width: 500
 
 
+.. _advanced-asymmetry-label:
 Advanced example using asymmetry comparison
 ===========================================
 
-This example shows how to plot an asymmetry plot between two histograms and two functions:
+This example shows how to plot an asymmetry plot between two histograms and two functions. Information on how to convert a function from different fitting library to an object that can be used by ``plothist`` can be found here :ref:`plot-fit-result-label`.
```
cyrraz, line 37: from different fitting library -> from an external fitting package

cyrraz (changes requested):

docs/usage/misc.rst
0ctagon: You are right, `n_data` is indeed the number of predicted data events in that range corresponding to a certain model component.

cyrraz (approved):

PR #92, opened 2023-12-14
added 2d example with discrete colormap

0ctagon:
Title. Some people might thing that the colormap is changed in `colorbar_kwargs`. Added documentation on `pcolormesh_kwargs` and a small complex example.


docs/examples/2d_hist/2d_hist_simple_discrete_colormap.py
```diff
@@ -0,0 +1,42 @@
...
+# We only take a subset of the data
+nentries = 1000
+
+h = make_2d_hist([df[name_x][:nentries], df[name_y][:nentries]], bins=[50, 50])
+
+from matplotlib.colors import ListedColormap
+from plothist import get_color_palette
+
+###
+# 0 entries will be white, the rest will have one color from the plasma colormap per entry value
```
cyrraz, line 27: The preferred way to have white for zero entry should be the use of the argument `vmin` of `ax.pcolormesh()`. Could you please try?

cyrraz (changes requested):

docs/examples/2d_hist/2d_hist_simple_discrete_colormap.py
0ctagon: It depends what we want.

We could do
```
cmap = ListedColormap(
    list(get_color_palette("plasma", int(h.values().max()) * 2 -1))
)

cmap.set_under("white")

fig, ax, ax_colorbar = plot_2d_hist(
    h, colorbar_kwargs={"label": "Entries"}, pcolormesh_kwargs={"cmap": cmap, "vmin": 0.5},
)
```

which gives:
image

but the white and 0 are not in the legend anymore. If you put `vmin` to 0, it will be colored.

I would prefer the 0 and white to be on the colorbar, that's why I proposed this solution.

cyrraz: OK

cyrraz (approved):

PR #93, opened 2023-12-14
fix mc_uncertainty plot for function logic error

0ctagon:
Before, if you only plot 1 function in `unstacked_component` of `plot_data_model_comparison`, the code wants to plot its unexisting `mc_uncertainty`. This PR fixes the bug.


cyrraz (approved):
Nice catch

PR #94, opened 2023-12-15
allow for non-square figsize when doing plot_2d_with_proj

0ctagon:
Title.
Sometimes you want your plot to reflect that the `x_range` is not the same as the `y_range`, and scale the figure accordingly. This PR allows the user to change the figsize to get this result:

Screenshot from 2023-12-15 10-02-04

### DIFF
--- a/plothist/plotters.py
+++ b/plothist/plotters.py
@@ -231,6 +231,7 @@ def plot_2d_hist_with_projections(
     pcolormesh_kwargs={},
     colorbar_kwargs={},
     plot_hist_kwargs={},
+    figsize=(6, 6),
 ):
     """Plot a 2D histogram with projections on the x and y axes.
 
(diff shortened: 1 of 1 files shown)


plothist/plotters.py
```diff
@@ -308,6 +310,7 @@ def plot_2d_hist_with_projections(
         ax_colorbar=ax_colorbar,
         pcolormesh_kwargs=pcolormesh_kwargs,
         colorbar_kwargs=colorbar_kwargs,
+        square_ax=False,
```
cyrraz, line 313: What about `square_ax=figsize[0]==figsize[1]`?

cyrraz (changes requested):

plothist/plotters.py
0ctagon: if `figsize[0]==figsize[1]`, this subplot is square shape by construction or the subplots placement, no need.

cyrraz: OK

cyrraz (approved):

PR #99, opened 2024-01-11
Bugfix: correctly overwrite the range in registry if specified

0ctagon:
Small bug, `overwrite=True` argument in `update_variable_registry_ranges()` was not working properly.


cyrraz (approved):

0ctagon:
Should we create a small release? Should we wait for other potential bugfixes/changes?

cyrraz:
> @cyrraz should we create a small release? Should we wait for other potential bugfixes/changes?

A small release does not hurt.

PR #102, opened 2024-01-24
Bugfix: Corner case in update_variable_registry_ranges

0ctagon:
The logic of the `variable_registry` is to get the data by using the `name` property and not the `variable_key`, which is not the case in the current `update_variable_registry_ranges` function.
This bug was found by a user playing with the registry and custom dictionary, and using the advanced registry example of the documentation

### DIFF
--- a/plothist/variable_registry.py
+++ b/plothist/variable_registry.py
@@ -266,7 +266,7 @@ def update_variable_registry_ranges(
     Raises
     ------
     RuntimeError
-        If the variable does not have a name, bins and/or range property in the registry.
+        If the variable does not have a name, bins or range property in the registry.
     """
     _check_if_variable_registry_exists(path)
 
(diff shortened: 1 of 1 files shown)


plothist/variable_registry.py
```diff
@@ -266,7 +266,7 @@ def update_variable_registry_ranges(
     Raises
     ------
     RuntimeError
-        If the variable does not have a bins or range property in the registry.
+        If the variable does not have a name, bins and/or range property in the registry.
```
cyrraz, line 269: I do not understand the and/or. The error is thrown if the variable is missing a name, bin or range field.

cyrraz (changes requested):

plothist/variable_registry.py
```diff
@@ -277,15 +277,15 @@ def update_variable_registry_ranges(
 
     for variable_key in variable_keys:
         variable = get_variable_from_registry(variable_key, path=path)
-        if "bins" not in variable.keys() or "range" not in variable.keys():
+        if not all(key in variable.keys() for key in ["bins", "range", "name"]):
             raise RuntimeError(
-                f"Variable {variable_key} does not have a bins or range property in the registry {path}."
+                f"Variable {variable_key} does not have a name, bins and/or range property in the registry {path}."
```
cyrraz, line 282: I do not understand the and/or. The error is thrown if the variable is missing a name, bin or range field.

cyrraz (changes requested):

cyrraz (approved):

PR #103, opened 2024-01-24
Renamed misc category

0ctagon:
This section being a bit larger than just a "misc" chapter, it's now a dedicated section.
Also could add `pyhf` example in this PR (and maybe `iminuit`).

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -29,7 +29,7 @@ Content
     basics/variable_registry
     usage/style
     usage/utilities
-    usage/misc
+    usage/plot_fitting_results
 
 .. toctree::
     :maxdepth: 2
--- a/docs/usage/plot_fitting_results.rst
+++ b/docs/usage/plot_fitting_results.rst
@@ -4,6 +4,67 @@
 Plot result of a fit
 ====================
 
+This section shows how to plot the result of a fit with ``pyhf``, ``RooFit`` or ``zfit``.
+
+pyhf
+****
+
+Here is an example of a fit with ``pyhf``, whose result is then shown using ``plothist``:
+
+.. code-block:: python
+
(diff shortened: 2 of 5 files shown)


docs/index.rst
```diff
@@ -29,7 +29,7 @@ Content
     basics/variable_registry
     usage/style
     usage/utilities
-    usage/misc
+    usage/plot_fitting_results
```
0ctagon, line 32: We can also put this under the `Documentation` chapter instead of the `Utilities`

docs/usage/plot_fitting_results.rst
```diff
@@ -1,22 +1,18 @@
...
-====
-Misc
-====
-
 .. _plot-fit-result-label:
-Plot fit results
-================
+
+====================
+Plot fitting results
```
cyrraz, line 4: Plot result of a fit, "fitting" sounds a but jargonic.

docs/index.rst
cyrraz: I think under `usage/` is still justified, because it is a specific usage that won't be used by every user.

cyrraz (changes requested):

docs/index.rst
0ctagon: I meant, in which category in the screenshot do we put it?

image
 
It is currently in `Utilities` under `misc`.

cyrraz: I would keep it in utilities

0ctagon:
@cyrraz Need new review.

I added a new RooFit example which should work for any PDFs shapes, without needing to understand the puzzling complicated un-fun hidden renormalization process that RooFit does when using the black-box `PlotOn` method.

PR #108, opened 2024-01-25
Add workflow for testing

cyrraz:
This PR add a workflow for running pytest when a PR is opened against the main branch or when we push to the main branch.

This tests the code in all the python versions ranking from 3.7 to 3.11.

As you can see below, the simple implemented test is passed in the five versions of Python.

### DIFF
--- a/tests/test_histogramming.py
+++ b/tests/test_histogramming.py
@@ -1,0 +1,15 @@
+from plothist import make_hist
+import boost_histogram as bh
+
+
+def test_make_hist():
+    """
+    Test make_hist() function.
+    """
+    h = make_hist(data=[0, 1, 2, 3, 4], bins=5, range=(0, 5))
+    assert isinstance(h, bh.Histogram)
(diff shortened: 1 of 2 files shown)


tests/test_histogramming.py
```diff
@@ -0,0 +1,15 @@
+from plothist import make_hist
+import boost_histogram as bh
+
+
+def test_make_hist():
```
0ctagon, line 5: Is this test only of simple example or an actual test we want to keep?
Should we add more tests in the PR?
Should we add something like `make_examples.py` in this PR?

cyrraz: * We can keep this simple test. Even if it is trivial, it is important that it works.
* More complex tests should be created in dedicated PRs, since some may take longer than others to work.
* The testing of the figure can take time to implement, I would not block the implementation of other tests because of that.

0ctagon (approved):

PR #110, opened 2024-01-29
Refactor plot_reordered_legend function to include optional legend location

0ctagon:
title

### DIFF
--- a/plothist/plothist_style.py
+++ b/plothist/plothist_style.py
@@ -355,7 +355,7 @@ def add_luminosity(
     )
 
 
-def plot_reordered_legend(ax, order, loc="best"):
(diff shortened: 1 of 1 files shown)


plothist/plothist_style.py
```diff
@@ -355,7 +355,7 @@ def add_luminosity(
     )
 
 
-def plot_reordered_legend(ax, order):
+def plot_reordered_legend(ax, order, loc="best"):
```
cyrraz, line 358: We should add `**kwargs` instead

cyrraz (changes requested):

cyrraz (approved):

PR #111, opened 2024-01-29
Time variable in italic in add_luminosity

0ctagon:
Variable should be in italic.
Also good to have an italic example in the doc because people are using by default `\textit{}` which doesn't work. Could add also a more explicit `\mathit{}` example in the doc.

### DIFF
--- a/plothist/plothist_style.py
+++ b/plothist/plothist_style.py
@@ -340,7 +340,7 @@ def add_luminosity(
         text += " "
     if is_data:
         if lumi:
-            text += rf"$\int\,\mathcal{{L}}\,\mathrm{{d\mathit{{t}}}}={lumi}\,{lumi_unit}^{{-1}}$"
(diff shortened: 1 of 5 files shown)


plothist/plothist_style.py
```diff
@@ -340,7 +340,7 @@ def add_luminosity(
         text += " "
     if is_data:
         if lumi:
-            text += rf"$\int\,\mathcal{{L}}\,\mathrm{{dt}}={lumi}\,{lumi_unit}^{{-1}}$"
+            text += rf"$\int\,\mathcal{{L}}\,\mathrm{{d\mathit{{t}}}}={lumi}\,{lumi_unit}^{{-1}}$"
```
cyrraz, line 343: Could you try instead `\mathrm{{d}}\mathit{{t}}`? This should give the same result, but the syntax is more logical.

cyrraz (changes requested):

cyrraz (approved):

PR #115, opened 2024-02-01
Support efficiency computation and documentation improvement

cyrraz:
* support efficiency computation.

The code is here, but update of the documentation is still needed.


plothist/comparison.py
```diff
@@ -201,38 +200,96 @@ def get_difference(h1, h2, h1_uncertainty_type="symmetrical"):
...
 
 
+def get_efficency(h1, h2):
+    """
+    Calculate the ratio of two correlated histograms (h1/h2), in which the entries of h1 are a subsample of the entries of h2.
+    The variances are calculated according to the formula given in :ref:`documentation-statistics-label`.
+
+    The following conditions must be fulfilled for the calculation of the efficiency:
+    * The bins of the histograms must be equal.
+    * The histograms must be unweighted.
```
0ctagon, line 240: What about the weighted case?

I think it has been implemented here: https://github.com/scikit-hep/hist/blob/ec2842326046bf860b12646222aa1d4ed45c2092/src/hist/intervals.py#L80

plothist/plotters.py
```diff
@@ -537,15 +537,21 @@ def plot_comparison(
...
-        How to treat the uncertainties of the histograms when comparison is "ratio" or "relative_difference" ("uncorrelated" for simple comparison, "split" for scaling and split h1 and h2 uncertainties). This argument has no effect if comparison != "ratio" or "relative_difference". Default is "uncorrelated".
+        How to treat the uncertainties of the histograms when comparison is "ratio" or "relative_difference":
+        This argument has no effect if comparison != "ratio" or "relative_difference".
+        The options are:
+        * "uncorrelated" for the comparison of two uncorrelated histograms,
+        * "split" for scaling the uncertainties of h1 by the inverse of the bin content of h2, i.e. assuming zero uncertainty coming from h2 in the ratio uncertainty, and plotting separately scaled h2 uncertainties.
+        Default is "uncorrelated".
     **plot_hist_kwargs : optional
         Arguments to be passed to plot_hist() or plot_error_hist(), called in case the comparison is "pull" or "ratio", respectively. In case of pull, the default arguments are histtype="stepfilled" and color="darkgrey". In case of ratio, the default argument is color="black".
```
0ctagon, line 556: Arguments to be passed to plot_hist() (case "pull") or plot_error_hist() (all other comparison types).

0ctagon (changes requested):
Beautiful PR.
Should we add a python test to check if `get_efficiency` returns the correct values for a simple example? or the future example is enough to check that?

plothist/comparison.py
cyrraz: I am not sure this covers the weighted case, this is just another approach. Since our approach is already less naïve than the most naïve one that is usually used: eff(1-eff)/N, so I would say this is a good compromise for now.

0ctagon:
Doc here: https://plothist.readthedocs.io/en/support-efficiency-computation/

docs/img/1d_comparison_efficiency.svg
0ctagon, line 1: `Efficiency` label is a bit large, but when scaled with `set_fitting_ylabel_fontsize`, it's a bit too small compared to `Entries` label.

docs/examples/1d_hist/1d_comparison_efficiency.py
```diff
@@ -0,0 +1,42 @@
...
+Efficiency comparison
+=====================
+
+Compare the ratio between two histograms h1 and h2 when the entries of h1 are a subset of the entries of h2.
+"""
+
+# Set the seed for the .sample() method
+import numpy as np
+
+np.random.seed(8311311)
```
cyrraz, line 11: This function is a legacy function and should not be used anymore.

* https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html

docs/examples/1d_hist/1d_comparison_efficiency.py
```diff
@@ -0,0 +1,42 @@
...
+
+from plothist import get_dummy_data
+
+df = get_dummy_data()
+
+name = "variable_1"
+
+x_total = df[name][df["category"] == 2]
+x_sample = x_total.sample(frac=0.75) # x_sample is a subset of x_total
```
cyrraz, line 20: Actually, this code will crash (!) if merged with the main: `df` is not a `pandas` DataFrame anymore, so you cannot use the `sample` method. I would recommend to just take the 75% of the array with a simple array slicing.

docs/examples/1d_hist/1d_comparison_only_efficiency.py
```diff
@@ -0,0 +1,37 @@
...
+np.random.seed(8311311)
+
+from plothist import get_dummy_data
+
+df = get_dummy_data()
+
+name = "variable_1"
+
+x_total = df[name][df["category"] == 2]
+x_sample = x_total.sample(frac=0.75) # x_sample is a subset of x_total
```
cyrraz, line 20: Actually, this code will crash (!) if merged with the main: `df` is not a `pandas` DataFrame anymore, so you cannot use the `sample` method. I would recommend to just take the 75% of the array with a simple array slicing.

docs/examples/1d_hist/1d_comparison_only_efficiency.py
```diff
@@ -0,0 +1,37 @@
...
+Only the comparison plot
+========================
+
+Example that shows how to only plot the comparison. Here, we compare the ratio between two histograms h1 and h2 when the entries of h1 are a subset of the entries of h2.
+"""
+
+# Set the seed for the .sample() method
+import numpy as np
+
+np.random.seed(8311311)
```
cyrraz, line 11: This function is a legacy function and should not be used anymore.

* https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html

cyrraz (commented):
Hello @0ctagon, I left a few comments in this review, please implement them. I cannot request changes since this is my own PR.

cyrraz:
> Beautiful PR. Should we add a python test to check if `get_efficiency` returns the correct values for a simple example? or the future example is enough to check that?

Before merging, I think it is a good idea to add tests:
* to check reasonable values of the efficiency, comparing with the simple formula `eff*(1-eff)/N`.
* to check that the errors are correctly raised when the efficiency is computed on invalid histograms.

cyrraz (commented):
@0ctagon Couple of new comments

tests/test_efficiency.py
```diff
@@ -16,12 +16,23 @@ def test_efficiency_subsample():
         assert True
 
 
+def simple_efficiency_uncertainty(total, sample):
```
cyrraz, line 19: This is derived from Binomial Statistics, please add it to the name of the function or to the docstring.

tests/test_efficiency.py
```diff
@@ -30,6 +41,31 @@ def test_efficiency_simple_values():
...
+    assert approx(values) == np.array([0.1])
+    assert approx(high_uncertainty) == np.array([9.48683493e-05])
+    assert approx(low_uncertainty) == np.array([9.48683493e-05])
+
+    assert approx(high_uncertainty[0]) == simple_efficiency_uncertainty(n1, n2) # 1e-6 relative error by default
+    assert approx(low_uncertainty[0]) == simple_efficiency_uncertainty(n1, n2) # 1e-6 relative error by default
+
+
+
+test_efficiency_simple_values()
```
cyrraz, line 67: Do we need to call the function for the test to be tried by pytest? I do not think so.

cyrraz (commented):
@0ctagon And a couple more.

tests/test_efficiency.py
0ctagon: oops

0ctagon (approved):
Efficiency correctly implemented

PR #117, opened 2024-02-06
Add troubleshooting instructions for installing Latin Modern fonts and Bugfix install_latin_modern_fonts.py script

0ctagon:
`install_latin_modern_fonts` was not correctly installing the fonts.
They were installed in `\~/.fonts/`, plus the unzip command was not working.
Added some documentation if the command is not recognize by the system (happened once).

### DIFF
--- a/plothist/scripts/install_latin_modern_fonts.py
+++ b/plothist/scripts/install_latin_modern_fonts.py
@@ -6,15 +6,19 @@ from pathlib import Path
 
 def install_latin_modern_fonts():
     """
-    Install Latin Modern fonts for matplotlib.
+    Install Latin Modern fonts for matplotlib and clean the font cache.
     """
     if platform.system() == "Linux":  # Linux
-        font_directory = Path("~/.fonts/")
+        font_directory = Path(os.path.expanduser("~/.fonts/"))
         font_directory.mkdir(parents=True, exist_ok=True)
-        matplotlib_font_cache = "~/.cache/matplotlib/fontlist-v330.json"
+        matplotlib_font_cache = Path(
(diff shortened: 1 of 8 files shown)


plothist/scripts/install_latin_modern_fonts.py
```diff
@@ -6,15 +6,19 @@
 
 def install_latin_modern_fonts():
     """
-    Install Latin Modern fonts for matplotlib.
+    Install Latin Modern fonts for matplotlib and clean the font cache.
```
cyrraz, line 9: clean -> clear

cyrraz (approved):
I would propose to add 1.0.5->1.0.6 version change to this PR and make a new release.
The efficiency feature will need a bit more time.

cyrraz (changes requested):
See comments on my previous review.

0ctagon:
@cyrraz  What do you think of removing the fonts section from `Style, fonts and colors` chapter and creating a `Fonts installation` chapter? The other one will only be `Style and colors`.

cyrraz:
> @cyrraz What do you think of removing the fonts section from `Style, fonts and colors` chapter and creating a `Fonts installation` chapter? The old chapter will be `Style and colors`. We can put this new chapter under `Utilities` or just below `Installation and update`

Yes, good idea.

cyrraz (changes requested):
Could you please try to add a pytest to check that the font can be installed? I this is complicated, we may leave it for later.

cyrraz (approved):

0ctagon:
I think this bugfix is too important to wait for a pytest to be implemented. The pytest will be added in a next PR.

PR #121, opened 2024-02-21
Update readme

cyrraz:
* Update README

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@
 plothist
 ========
 
-**Plot histograms in a scalable way and a beautiful style.**
+**Visualize data in a scalable way and a beautiful style.**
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -3,7 +3,7 @@
 plothist
 ========
 
-**Plot histograms in a scalable way and a beautiful style.**
+**Visualize data in a scalable way and a beautiful style.**
```
0ctagon, line 6: It can do more than just visualize, but I think this verb is broad enough.

0ctagon (approved):

README.rst
cyrraz: Yes, I wanted to stay short. What about Visualize and compare...?

0ctagon: Agreed.

PR #123, opened 2024-02-27
Added test for font installation

0ctagon:
This PR took more time than expected, was a pain to debug, but in the end it works :rocket: 
Sometimes during the pytest, the font URL can be unreachable, thus making the test fail. Reloading the test fixes the problem.


tests/test_install_fonts.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            new_fontmanager.findfont(
+                f"Latin Modern {font_type}", fallback_to_default=False
+            )
+        except ValueError:
+            print(f"The font Latin Modern {font_type} was not found.")
+            installation_failed = True
+
+    if installation_failed:
+        fail(f"One of the font was not installed correctly.")
```
cyrraz, line 28: At least one of the Latin Modern fonts was not installed correctly.

cyrraz (changes requested):
Very nice PR :)

tests/test_install_fonts.py
```diff
@@ -0,0 +1,28 @@
+from pytest import fail
+from plothist.scripts import install_latin_modern_fonts
+
+
+def test_install_latin_modern_fonts():
+    """
+    Test install_latin_modern_fonts.
+    """
+
+    install_latin_modern_fonts()
```
cyrraz, line 10: Could you please try the option `--wait=x` with x=2, 5, 10 to the `wget` command in the installation script to see whether the test becomes more stable in that case in case the url is not immediately reachable?

cyrraz (changes requested):

cyrraz (approved):

PR #126, opened 2024-02-28
added a note to show how to plot hist unc as hashed area

0ctagon:
* Added a note to show how to plot hist unc as hashed area.

### DIFF
--- a/plothist/plotters.py
+++ b/plothist/plotters.py
@@ -392,7 +392,7 @@ def plot_error_hist(hist, ax, uncertainty_type="symmetrical", density=False, **k
 
 def plot_hist_uncertainties(hist, ax, **kwargs):
     """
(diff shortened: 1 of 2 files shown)


plothist/plotters.py
```diff
@@ -392,7 +392,7 @@ def plot_error_hist(hist, ax, uncertainty_type="symmetrical", density=False, **k
 
 def plot_hist_uncertainties(hist, ax, **kwargs):
     """
-    Plot the uncertainties of a histogram as a hatched area.
+    Plot the symmetrical uncertainty of a histogram, the Poisson standard deviation derived from the variance stored in the histogram, as a hatched area.
```
cyrraz, line 395: which is the Poisson ...

cyrraz (changes requested):

cyrraz (approved):

PR #130, opened 2024-02-29
added fix if certificate not found for font website

0ctagon:
Should work better now.

### DIFF
--- a/src/plothist/scripts/install_latin_modern_fonts.py
+++ b/src/plothist/scripts/install_latin_modern_fonts.py
@@ -35,20 +35,27 @@ def install_latin_modern_fonts():
         )
 
     # Install Latin Modern Math
-    subprocess.run(
-        [
-            "wget",
-            "--retry-connrefused",  # retry refused connections and similar fatal errors
-            "--retry-on-host-error",# retry on host errors such as 404 "Not Found"
-            "--waitretry=1",        # wait 1 second before next retry
-            "--read-timeout=20",    # wait a maximum of 20 seconds in case no data is received and then try again
-            "--timeout=15",         # wait max 15 seconds before the initial connection times out
-            "-t", "10",             # retry 10 times
-            "-P",
-            font_directory,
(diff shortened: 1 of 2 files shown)


src/plothist/scripts/install_latin_modern_fonts.py
```diff
@@ -35,20 +35,27 @@ def install_latin_modern_fonts():
...
+                    "-P",
+                    font_directory,
+                    "http://mirrors.ctan.org/fonts/lm-math/opentype/latinmodern-math.otf",
+                ],
+                check=True
+            )
+            break
+        except subprocess.CalledProcessError:
+            time.sleep(5)
+
```
cyrraz, line 58: * Why only this font download? Shouldn't we apply the patch to the other font downloads?
* I think the following version is a little bit more elegant by avoiding the `try-except` and `break` structure (we could also make it a function to avoid to repeat it for each font):

```
command = [
    "wget",
    "--retry-connrefused",  # retry refused connections and similar fatal errors
    "--retry-on-host-error",  # retry on host errors such as 404 "Not Found"
    "--waitretry=1",  # wait 1 second before next retry
    "--read-timeout=20",  # wait a maximum of 20 seconds in case no data is received and then try again
    "--timeout=15",  # wait max 15 seconds before the initial connection times out
    "-t",
    "10",  # retry 10 times
    "-P",
    font_directory,
    "http://mirrors.ctan.org/fonts/lm-math/opentype/latinmodern-math.otf",
]

attempt = 0
max_attempt = 10
success = False

while not success and attempt < max_attempt:
    result = subprocess.run(command, capture_output=True, text=True)
    success = result.returncode == 0
    if not success:
        # Increment attempt counter and wait before the next attempt
        attempt += 1
        time.sleep(5)

if not success:
    raise RuntimeError(f"Failed to download after {max_attempt} attempts.")

```

cyrraz (changes requested):
Nice :)

cyrraz (approved):

PR #132, opened 2024-03-05
Implementation of the pytest to recreate the examples in the doc

0ctagon:
@cyrraz this PR implements a pytest that automatically checks if any .svg files have been modified by the lastest commit. 

Commit [16915a3](https://github.com/cyrraz/plothist/pull/132/commits/16915a33643de7d5b84ad0c4dd9a0434426071c5) is the first one with a proper implementation. 
Commit [4cd86c5](https://github.com/cyrraz/plothist/pull/132/commits/4cd86c5af5657b9618f2cb2ac9c0d045469e142e) was a successful test to see if it fails if some .svg are modified.
In addition, removing the test that installs the fonts also causes the test fail. 

If no .svg changes, the test always passes for Python 3.9, 3.10 and 3.11.
Python 3.7 always fails, and Python 3.8 only fails for some plots, not all.
I haven't looked into why they fail at the moment, but I'm random guessing some seed/metadata changes with python/matplotlib and svg files. Let's discuss.

It also implements a new `plothist_make_examples` command (that can be run from anywhere), which is equivalent to running `python3 make_examples` in `plothist/src/scripts` (the yaml metadata is downloaded from github [here](https://github.com/0ctagon/plothist-utils)).


0ctagon:
@cyrraz after some investigation, it was indeed the behaviour of svg files that changed with the python version.

This commit [a19a78d](https://github.com/0ctagon/plothist-test/pull/1/commits/a19a78dde6958069c6a3f683939ccf4b0d806f4f) shows the images that changed with python3.8, and this commit [855e8cc](https://github.com/0ctagon/plothist-test/pull/1/commits/855e8ccaef36b173cb1e0b9ba613fa6f6e1ecdbd) with python3.7.
As we can see, nothing changes for the images, but the encoding of the svg files changes a bit.
We can also see that the respective pytests for the matching python version passed without problems, while the other one crashed.

To fix this, I added a small check for the python version at the beginning of the script, so that it only produces images with python3.9+.

src/plothist/scripts/make_examples.py
```diff
@@ -1,104 +1,191 @@
...
-            ]
+import plothist
+from pytest import fail
+import hashlib
+import matplotlib.pyplot as plt
+import warnings
+import sys
+
+# Set figure.max_open_warning to a large number to avoid warnings
+plt.rcParams["figure.max_open_warning"] = 1000
```
cyrraz, line 12: Shouldn't we close the figures once they are saved to avoid the warning instead of muting it?

src/plothist/scripts/make_examples.py
```diff
@@ -1,104 +1,191 @@
...
+
+
+def make_examples(no_input=False, check_svg=False, print_code=False):
+    """
+    This function cam redo automatically all the examples from the documentation. Only works with python 3.9 or higher.
```
cyrraz, line 17: cam -> can

cyrraz (changes requested):
Nice PRs, two minor comments.

src/plothist/scripts/make_examples.py
0ctagon: Nice catch, it's an old artifact from the old method, no need now.

cyrraz (approved):

PR #138, opened 2024-03-21
Turn off under/overflow bin by default, making the upper edge inclusive instead of exclusive

0ctagon:
Outdated comment, this PR changed direction during review (title was updated):

> From `max(data)` to `max(data)+((max(data)-min(data))/(bins-1))`.
> 
> With a simple histogram:
> `h = make_hist([0]*3+[1]*4+[2]*5+[3]*6+[4]*7+[5]*8, bins=6)`
> 
> Before:
> Screenshot from 2024-03-21 15-30-41
> `UserWarning: Only 75.76% of data contained in the binning range [0.0, 5.0). Note that the upper edge is exclusive.`
> 
> After:
> Screenshot from 2024-03-21 15-30-54
> `No warnings`


src/plothist/histogramming.py
```diff
@@ -51,7 +51,7 @@ def create_axis(bins, range=None, data=np.array([])):
                 "Please supply a range or provide data."
             )
         x_min = min(data) if range[0] == "min" else range[0]
-        x_max = max(data) if range[1] == "max" else range[1]
+        x_max = max(data)+((max(data)-min(data))/(bins-1)) if range[1] == "max" else range[1]
```
cyrraz, line 54: Need to consider the cases bins<=0 and bins=1

src/plothist/histogramming.py
```diff
@@ -62,7 +62,7 @@ def create_axis(bins, range=None, data=np.array([])):
         # handle empty arrays. Can't determine range, so use 0-1.
         x_min, x_max = 0, 1
     else:
-        x_min, x_max = min(data), max(data)
+        x_min, x_max = min(data), max(data)+((max(data)-min(data))/(bins-1))
```
cyrraz, line 65: Need to consider the cases bins<=0 and bins=1

src/plothist/histogramming.py
cyrraz: Here, what I do not like, it that we do not apply what the user asks us to do. The user want the upper range to be max(data) and we do not do that. We should mention it explicitly in the documentation of the function and also in the readthedocs. I propose you put the example you show in the summary of this PR in the readthedoc to explain the "exclusive upper-edge" situation.

cyrraz (commented):
Nice PR, that should diminish the probability of mistake from the user.

Could you please check why the warning does not appear on Jupyter Notebook?

src/plothist/histogramming.py
0ctagon: "you put the example you show in the summary of this PR in the readthedoc to explain the "exclusive upper-edge" situation."
In which section? I think the docstring is enough.

src/plothist/histogramming.py
```diff
@@ -13,7 +21,7 @@ def create_axis(bins, range=None, data=np.array([])):
     bins : int or array-like
         The number of bins or bin edges for the axis.
     range : None or tuple, optional
-        The range of the axis. If None, it will be determined based on the data.
+        The range of the axis. If None or ["min", "max"], it will return the interval [min(data), max(data)+bin_width] if bins > 1, otherwise [min(data), max(data)]. The additional bin width is added to the max value to avoid data points falling on the upper edge of the last bin, which is exclusive for boost-histogram.
```
cyrraz, line 24: * If the user wants a histogram with 1 bins, it is likely to count the total number of entries, so we should not use [min(data), max(data)] which will miss the entries as max(data). What about [min(data), max(data)+abs(max(data)-min(data))] if bins=1? It gives an intuitive range of [0, 2) in case of a binary variable taking either 0 or 1 values.
* falling on -> located
* for boost-histogram -> in boost_histogram
* The case is so specific that I move it at the end of the doc line, starting a new sentence with something like "In the special case bins=1, ..."

src/plothist/histogramming.py
```diff
@@ -51,7 +64,13 @@ def create_axis(bins, range=None, data=np.array([])):
                 "Please supply a range or provide data."
             )
         x_min = min(data) if range[0] == "min" else range[0]
-        x_max = max(data) if range[1] == "max" else range[1]
+        if range[1] == "max":
+            if bins == 1:
+                x_max = max(data)
```
cyrraz, line 69: see my other comment.

src/plothist/histogramming.py
```diff
@@ -1,9 +1,17 @@
...
 import warnings
 
 
+# Define a custom warning for range issues
+class RangeWarning(Warning):
+    pass
+
+
+# Always show the range warnings
+warnings.filterwarnings("always", category=RangeWarning)
```
cyrraz, line 12: Nice :)

cyrraz (changes requested):
I realize that I am peaky for this PR, but this is indeed a very important point in plothist. I think we need to reconsider our global approach here, maybe by making the upper-edge inclusive in case of a regular axis, as boost_histogram kind of does it. Let's discuss it at a meeting.

0ctagon:
Now works like matplotib

Screenshot from 2024-03-25 19-07-45

src/plothist/__init__.py
```diff
@@ -116,3 +116,11 @@
             stacklevel=3,
         )
         break
+
+# Check version of boost_histogram
+import boost_histogram as bh
+
+if tuple(int(part) for part in bh.__version__.split(".")) < (1, 4, 0):
+    raise ImportError(
+        "The version of boost_histogram is lower than 1.4.0. Please update to the latest version to avoid issues (pip install --upgrade boost_histogram==1.4.0).",
```
cyrraz, line 125: Remove "==1.4.0", I think we aim at always be up-to-date with the latest version of bh, so potentially >1.4.0

src/plothist/histogramming.py
```diff
@@ -16,6 +24,10 @@ def create_axis(bins, range=None, data=np.array([])):
         The range of the axis. If None, it will be determined based on the data.
     data : array-like, optional
         The input data for determining the axis range. Default is an empty array.
+    overflow : bool, optional
+        Whether to include overflow bins. If False, the upper edge of the last bin is inclusive. Default is False.
```
cyrraz, line 28: "an overflow bin", there is only one bin.

src/plothist/histogramming.py
```diff
@@ -16,6 +24,10 @@ def create_axis(bins, range=None, data=np.array([])):
...
+    underflow : bool, optional
+        Whether to include underflow bins. Default is False.
```
cyrraz, line 30: "an underflow bin", there is only one bin.

src/plothist/histogramming.py
```diff
@@ -41,7 +55,10 @@ def create_axis(bins, range=None, data=np.array([])):
     if N > 1:
         if range is not None:
             warnings.warn(f"Custom binning -> ignore supplied range ({range}).")
-        return bh.axis.Variable(bins)
+        return bh.axis.Variable(bins, overflow=overflow, underflow=underflow)
```
cyrraz, line 58: underflow before afterflow

src/plothist/histogramming.py
```diff
@@ -71,7 +89,7 @@ def create_axis(bins, range=None, data=np.array([])):
         x_min = x_min - 0.5
         x_max = x_max + 0.5
 
-    return bh.axis.Regular(bins, x_min, x_max)
+    return bh.axis.Regular(bins, x_min, x_max, overflow=overflow, underflow=underflow)
```
cyrraz, line 92: underflow before afterflow

src/plothist/histogramming.py
```diff
@@ -108,12 +131,20 @@ def make_hist(data=np.array([]), bins=50, range=None, weights=1):
...
         h.fill(data, weight=weights, threads=0)
 
-        # Check what proportion of the data is in the underflow and overflow bins
-        range_coverage = h.values().sum() / h.values(flow=True).sum()
-        # Issue a warning in more than 1% of the data is outside of the binning range
+        # Check what proportion of the data outside of the binning range
+        if isinstance(weights, (int, float)):
+            n_data = len(data)*weights
+        else:
+            n_data = np.sum(np.array(weights))
```
cyrraz, line 138: `np.as_array(weights)` avoids extra copy I think

src/plothist/histogramming.py
```diff
@@ -168,12 +204,20 @@ def make_2d_hist(data=np.array([[], []]), bins=(10, 10), range=(None, None), wei
...
         h.fill(*data, weight=weights, threads=0)
 
-        # Check what proportion of the data is in the underflow and overflow bins
-        range_coverage = h.values().sum() / h.values(flow=True).sum()
-        # Issue a warning in more than 1% of the data is outside of the binning range
+        # Check what proportion of the data outside of the binning range
+        if isinstance(weights, (int, float)):
+            n_data = len(data[0]) * weights
+        else:
+            n_data = np.sum(np.array(weights))
```
cyrraz, line 211: `np.as_array(weights)` avoids extra copy I think

cyrraz (changes requested):

cyrraz:
Very useful PR, inclusive upper edge should make mistakes less likely. I will add a couple of pytest to this PR.

src/plothist/histogramming.py
0ctagon: afterflow :thinking:

src/plothist/histogramming.py
0ctagon: afterflow :thinking:

tests/test_histogramming.py
```diff
@@ -13,3 +16,82 @@ def test_make_hist():
...
+
+    data = [0] * 3 + [1] * 4 + [2] * 5 + [3] * 6 + [4] * 7 + [5] * 8
+    h1 = make_hist(data=data * 2, bins=5, range=(0, 5))
+    h2 = make_hist(data=data * 2, bins=[0, 1, 2, 3, 4, 5])
+    h3 = make_hist(data=data, bins=[0, 1, 2, 3, 4, 5], weights=2)
+    h4 = make_hist(data=data, bins=[0, 1, 2, 3, 4, 5], weights=[2] * len(data))
+
+    expected = [6, 8, 10, 12, 30]
+
+    assert np.array_equal(h1.values(), expected)
```
0ctagon, line 35: Why not a for loop?

tests/test_histogramming.py
```diff
@@ -13,3 +16,82 @@ def test_make_hist():
...
+    h5.fill(data * 2)
+    h6.fill(data, weight=[2] * len(data))
+    h7.fill(data, weight=[2] * len(data))
+    h8.fill(data, weight=[2] * len(data))
+    h9.fill(data, weight=[2] * len(data))
+    h10.fill(data, weight=[2] * len(data))
+
+    expected = [6, 8, 10, 12, 14]
+
+    assert np.array_equal(h1.values(), expected)
```
0ctagon, line 79: Why not a for loop?

tests/test_histogramming.py
```diff
@@ -13,3 +16,82 @@ def test_make_hist():
...
+    assert np.array_equal(h4.values(), expected)
+    assert np.array_equal(h5.values(), expected)
+    assert np.array_equal(h6.values(), expected)
+    assert np.array_equal(h7.values(), expected)
+    assert np.array_equal(h8.values(), expected)
+    assert np.array_equal(h9.values(), expected)
+    assert np.array_equal(h10.values(), expected)
+
+
+def test_range_coverage_warning():
```
0ctagon, line 91: Could we also have a check with a weighted histogram (two cases: a simple `int` and a `list`)?

0ctagon:
When PR is done, we can also push the `1.1.2 -> 1.1.3 or 1.2` change,

cyrraz (approved):
Approved, but please have another read at the whole PR to double check, this is an important PR.

tests/test_histogramming.py
```diff
@@ -13,3 +16,80 @@ def test_make_hist():
...
+    h9.fill(data, weight=[2] * len(data))
+    h10.fill(data, weight=[2] * len(data))
+
+    expected = [6, 8, 10, 12, 14]
+
+    for h in [h1, h2, h3, h4, h5, h6, h7, h8, h9, h10]:
+        assert np.array_equal(h.values(), expected)
+
+
+def test_range_coverage_warning():
```
0ctagon, line 81: Should we also test 2d hists?

src/plothist/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Plot histograms in a scalable way and a beautiful style."""
 
-__version__ = "1.1.2"
+__version__ = "1.2.0"
```
0ctagon, line 3: In the past we were doing `1.X` not `1.X.0`. I don't mind if we change now.

cyrraz: Let's change then, easier to find+replace

tests/test_histogramming.py
cyrraz: done

PR #139, opened 2024-03-26
For 1d simple hists: changed from c2 and c3 label to h1 and h2

0ctagon:
Figures will be updated after #138 is merged.


docs/img/1d_elt2.svg
cyrraz, line 1: Is it wanted to simplify this example? Is there any line in the doc that needs to be updated to accommodate this change?

cyrraz (changes requested):

docs/img/1d_elt2.svg
0ctagon: After looking at what we did for `plot_function` (only 1 simple function), I wanted the same for `plot_error_hist`. The old plot was too crowded, the new one only focus on `plot_error_hist`.
The doc around it doesn't need update.

cyrraz (approved):

cyrraz (approved):

PR #141, opened 2024-03-27
Fonts bugfix

0ctagon:
The `install_latin_modern_fonts` crashes if `wget` is too old, and it appears that `wget` is indeed old on some computing centers...
Also, `pytest` is used in the `install_latin_modern_fonts` and `plothist_make_examples`, so we should add it as a dependency -> Not needed anymore

### DIFF
--- a/src/plothist/scripts/make_examples.py
+++ b/src/plothist/scripts/make_examples.py
@@ -2,7 +2,6 @@ import os
 import yaml
 import subprocess
 import plothist
-from pytest import fail
 import hashlib
 import warnings
 import sys
@@ -34,6 +33,9 @@ def make_examples(no_input=False, check_svg=False, print_code=False):
             stacklevel=2,
         )
         return 1
+
+    if check_svg:
+        from pytest import fail
 
     plothist_folder = (
         plothist.__path__[0]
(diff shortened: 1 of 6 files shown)


src/plothist/scripts/make_examples.py
```diff
@@ -35,6 +34,9 @@ def make_examples(no_input=False, check_svg=False, print_code=False):
         )
         return 1
 
+    if check_svg:
+        from pytest import fail
```
cyrraz, line 38: Could be moved just before the use on line 155

cyrraz (changes requested):
Please also add 1.2.0 -> 1.2.1

cyrraz (approved):

cyrraz:
I approved the PR, but please consider my two comments above.

0ctagon:
The new unzip checks in the install font script fix this error I got recently:

```
STDOUT: Archive:  /path/.fonts/latin-modern-roman.zip

STDERR:   End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of /path/.fonts/latin-modern-roman.zip or
       /path/.fonts/latin-modern-roman.zip.zip, and cannot find/path/.fonts/latin-modern-roman.zip.ZIP, period.
```

PR #144, opened 2024-03-30
Update issue templates

cyrraz:
Add issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/user-story.md
+++ b/.github/ISSUE_TEMPLATE/user-story.md
@@ -1,0 +1,23 @@
+---
+name: User Story
+about: This template is for creating user stories.
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**As a** [role]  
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/user-story.md
0ctagon, line 1: What will this looks like? Could you provide an example?
I don' really get the 
```
**As a** [role]  
**I need** [function]  
**So that** [benefit]  
```
it's really oriented towards asking for features, but what if the issue is bug related?

cyrraz: As a user
I need plothist to be able to find the LM fonts if they are already present on my system
So that I don't have to reinstall the fonts.

This template is coming from a course about Agile/Scrum development by IBM, I think it makes sense. We can discuss it further at a meeting if you want.

0ctagon (approved):

PR #145, opened 2024-04-01
Added variable registry pytests

0ctagon:
(maybe) the last pytest we were missing


cyrraz:
> (maybe) the last pytest we were missing

Nothing lasts forever.

PR #146, opened 2024-04-02
Added pre-commit to the github workflow

0ctagon:
title


0ctagon:
Python 3.7 crashes because black==24.2.0 is not available. @cyrraz Do you remember why you chose this version of black? Should we downgrade black or remove the Python3.7 test?

cyrraz:
> Python 3.7 crashes because black==24.2.0 is not available. @cyrraz Do you remember why you chose this version of black? Should we downgrade black or remove the Python3.7 test?

I think we should still support Python 3.7, version 24.2.0 of `black` was chosen because it was the latest `black` version at the time we merged this code. There is an automatic command in `precommit` to update the repositories to their latest version.

Could we remove only the precommit test for Python 3.7?

cyrraz (approved):
Nice.

PR #150, opened 2024-04-19
update to v1.2.2

cyrraz:
* update to v1.2.2


0ctagon:
I also want to add/remove new badges with this PR.
Could we:

- Remove doc vX.X.X
- Add zenodo

0ctagon (approved):

PR #152, opened 2024-04-19
Add requirements for readthedoc to build

0ctagon:
readthedoc changed some internal logic, so it was not building.
The PR fixes the problem, see https://plothist.readthedocs.io/en/documentation-build-fix/


.readthedocs.yaml
```diff
@@ -8,8 +8,13 @@ version: 2
...
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
 
 # Build documentation in the "docs/" directory with Sphinx
 sphinx:
   configuration: docs/conf.py
+
+# Explicitly set the version of Python and its requirements
```
cyrraz, line 17: The comment is not correct. Remove or replace by "Install required packages" or similar

cyrraz (approved):
Thank you, just a trivial comment.

.readthedocs.yaml
0ctagon: The comment comes directly from readthedoc doc: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#id1

PR #154, opened 2024-06-04
Add text upgrade

0ctagon:
Due to popular demand, more aliases have been added to the ``add_text`` function.
Only the ``bottom`` option is not retro-compatible, but it was really useless with its old value. `Left`, `right` and `top` values are the same as before.

Documentation to see the table correctly here: https://plothist.readthedocs.io/en/add_text_upgrade/usage/utilities.html#add-text

PR #155, opened 2024-06-05
Add plot_only option in plot_data_model_comparison

0ctagon:
Only getting the main or comparison ax from `plot_data_model_comparison()` is not that trivial. Copying/moving axes to other figures is impossible to do in matplotlib. To redo the axes, the user could redo the plot by calling `plot_model()` and `plot_error_hist()`, or copy/pasting what we do for functions for the comparison ax, but this solution is complicated and the plot is already here in `plot_data_model_comparison()`. The user could use `fig.delaxes()`, but he needs to put the ticks and label back for `ax_main`, and the `figzise` will still be the 2 axes one, resizing the `ax_comparison` to a standard size is messy.`

This proposition is a small quality of life improvement to give the user the possibility to only keep one of the two axes on the fig, and remove the other, that works cleanly.

model_only_main
model_only_comp

### DIFF
--- a/src/plothist/plotters.py
+++ b/src/plothist/plotters.py
@@ -995,10 +995,10 @@ def plot_data_model_comparison(
     model_uncertainty=True,
     model_uncertainty_label="Model stat. unc.",
     data_uncertainty_type="asymmetrical",
-    plot_only=None,
     fig=None,
     ax_main=None,
     ax_comparison=None,
+    plot_only=None,
     **comparison_kwargs,
 ):
     """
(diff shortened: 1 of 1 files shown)


src/plothist/plotters.py
```diff
@@ -1080,11 +1083,25 @@ def plot_data_model_comparison(
             _check_counting_histogram(component)
 
     if fig is None and ax_main is None and ax_comparison is None:
-        fig, (ax_main, ax_comparison) = create_comparison_figure()
+        if plot_only is None:
+            fig, (ax_main, ax_comparison) = create_comparison_figure()
+        elif plot_only == "ax_main":
+            fig, ax_main = plt.subplots()
+            fig_temp, ax_comparison = plt.subplots()
```
cyrraz, line 1090: fig_temp -> _

src/plothist/plotters.py
```diff
@@ -1080,11 +1083,25 @@ def plot_data_model_comparison(
...
+        elif plot_only == "ax_comparison":
+            fig, ax_comparison = plt.subplots()
+            fig_temp, ax_main = plt.subplots()
```
cyrraz, line 1093: fig_temp -> _

src/plothist/plotters.py
```diff
@@ -1040,6 +1041,8 @@ def plot_data_model_comparison(
         The label for the model uncertainties. Default is "Model stat. unc.".
     data_uncertainty_type : str, optional
         What kind of bin uncertainty to use for data_hist: "symmetrical" for the Poisson standard deviation derived from the variance stored in the histogram object, "asymmetrical" for asymmetrical uncertainties based on a Poisson confidence interval. Default is "asymmetrical".
+    plot_only : str, optional
```
cyrraz, line 1044: I would put this argument after fig, ax_main, ax_comparison, so that these are defined and the explanation can be understood. We could also add that this argument can only be used when fig, ax_main and ax_comparison are not provided by the user.

src/plothist/plotters.py
```diff
@@ -995,6 +995,7 @@ def plot_data_model_comparison(
     model_uncertainty=True,
     model_uncertainty_label="Model stat. unc.",
     data_uncertainty_type="asymmetrical",
+    plot_only=None,
```
cyrraz, line 998: I would put this argument after fig, ax_main, ax_comparison, so that these are defined and the explanation can be understood. We could also add that this argument can only be used when fig, ax_main and ax_comparison are not provided by the user.

cyrraz (changes requested):
Nice, thank you for this PR.

0ctagon:
We could do a minor release after #154 and this PR are merged

cyrraz (approved):

PR #159, opened 2024-06-15
Minor text changes

cyrraz:
A set of textual changes to improve the example gallery.


0ctagon:
Doc here: https://plothist.readthedocs.io/en/minor-text-changes/

docs/examples/model_ex/model_all_comparisons.py
```diff
@@ -1,8 +1,8 @@
 """
-All data/model comparisons
-==========================
+Data/model comparisons
+======================
 
-This example shows every possible comparison between data and model.
+Multiple comparisons between data and model.
```
0ctagon, line 5: Why omit the fact that it is indeed all the possible comparisons that make sense with `plot_data_model_comparison()`?

docs/examples/model_ex/model_all_comparisons_no_model_unc.py
```diff
@@ -1,8 +1,8 @@
 """
-All data/model comparisons, no model uncertainty
-================================================
+Data/model comparisons, no model uncertainty
+============================================
 
-This example shows every possible comparison between data and model, without model uncertainty.
+Multiple comparisons between data and model, without model uncertainty.
```
0ctagon, line 5: Why omit the fact that it is indeed all the possible comparisons that make sense with `plot_data_model_comparison()`?

docs/examples/model_ex/model_examples_pull.py
```diff
@@ -2,7 +2,7 @@
 Pull plot
```
0ctagon, line 2: Data vs Model, pull comparison

docs/examples/model_ex/model_examples_pull_no_model_unc.py
```diff
@@ -2,7 +2,7 @@
 Pull plot, no model uncertainty
```
0ctagon, line 2: Data vs Model, no model uncertainty

docs/examples/model_ex/ratio_data_vs_model_with_stacked_and_unstacked_function_components.py
```diff
@@ -1,8 +1,8 @@
 """
-Model with data, stacked and unstacked functions
-================================================
+Compare data and functional model
```
0ctagon, line 2: Data vs functional model [to stay consistent with the Data vs Model notation]

docs/examples/model_ex/model_examples_pull_no_model_unc.py
```diff
@@ -58,6 +58,6 @@
     model_uncertainty=False,  # <--
 )
 
-add_luminosity(collaboration="Beast III", ax=ax_main, is_data=False)
+add_luminosity(collaboration="Plothist", ax=ax_main, is_data=False)
```
0ctagon, line 61: I liked showing a space in the example for the collaboration, as people might do unnecessary `\,`  or `$$` modes.

docs/examples/model_ex/model_examples_stacked.py
```diff
@@ -67,6 +67,8 @@
 
 ax_main.legend()
 
-add_luminosity(collaboration="Beast III", ax=ax_main, lumi=200, preliminary=True)
+add_luminosity(
+    collaboration="Plothist", ax=ax_main, lumi=3, lumi_unit="zb", preliminary=True
```
0ctagon, line 71: I liked showing a space in the example for the collaboration, as people might do unnecessary `\,`  or `$$` modes.

0ctagon (approved):
Nice PR.
Please check the few not that important comments before merging.

pyproject.toml
```diff
@@ -24,7 +24,7 @@ requires-python = ">=3.7"
 dependencies = [
     "boost-histogram~=1.4.0",
     "numpy>=1.14.5",
-    "matplotlib>=3.0",
+    "matplotlib<=3.8.1",
```
0ctagon, line 27: Adding
```

      run: |
        python -m pip install --upgrade pip
        python -m pip install matplotlib==3.8.1
        python -m pip install flit pytest pre-commit
        flit install
 ```
 in `python-package.yaml` instead

File: docs/examples/model_ex/model_examples_pull.py (hunk 3 above)
Comment (cyrraz) [reply docs/examples/model_ex/model_examples_pull.py:2]: Since we are already in the data vs model section, I prefer to keep it simpler.

File: docs/examples/model_ex/model_examples_pull_no_model_unc.py (hunk 6 above)
Comment (0ctagon) [reply docs/examples/model_ex/model_examples_pull_no_model_unc.py:61]: Also do we want upper case `Plohist`? aren't we lowercase `plothist`?

File: docs/examples/model_ex/model_examples_pull_no_model_unc.py (hunk 4 above)
Comment (cyrraz) [reply docs/examples/model_ex/model_examples_pull_no_model_unc.py:2]: Since we are already in the data vs model section, I prefer to keep it simpler.

File: docs/examples/model_ex/model_examples_pull_no_model_unc.py (hunk 6 above)
Comment (cyrraz) [reply docs/examples/model_ex/model_examples_pull_no_model_unc.py:61]: I changed to lower case. For the space, not sure it is necessary, as a user, I would naively assume that I do not need \,

File: pyproject.toml (hunk 8 above)
Comment (0ctagon) [reply pyproject.toml]:
```
    python -m pip install matplotlib==3.8.1 || python -m pip install matplotlib>3.0
```

PR #163, opened 2024-06-19
latest matplotlib

cyrraz:


0ctagon:
Only the hash changed, even if we have `svg_hashsalt` in our style. Is there other hash we need to set?

The culprit lies in one of this package:
```
Successfully installed contourpy-1.2.1 fonttools-4.53.0 numpy-2.0.0 pillow-10.3.0 pyparsing-3.1.2 python_dateutil-2.9.0.post0 scipy-1.13.1
```

0ctagon (commented):
We need to make sure that `plothist_make_examples` creates the correct plots, right now it only has a lower limit on the matplotlib version, which we saw is not enough.

Edit: we should also monitor `numpy` version, which has an impact on the hash of the saved svg. Adding the same threshold as matplotlib might be a good solution,

cyrraz:
> We need to make sure that `plothist_make_examples` creates the correct plots, right now it only has a lower limit on the matplotlib version, which we saw is not enough.
> 
> Edit: we should also monitor `numpy` version, which has an impact on the hash of the saved svg. Adding the same threshold as matplotlib might be a good solution.
> 
> Note: I had `numpy-1.26.4` because tensorflow doesn't like higher versions. Which version should we settle on?

A new major release of numpy/matplotlib is rare, so I think we do not need to take additional measures, but let's discuss in a meeting.

0ctagon (approved):
Excellent.

PR #170, opened 2024-06-27
Increased line length for blacken docs

0ctagon:
For better readability, only increased maximum line length for the doc


cyrraz (approved):
Good idea.

PR #177, opened 2024-10-07
feat: compatibility with mplhep

andrzejnovak:
This is still a draft, but I am opening it now to get CI running...


0ctagon:
In the end I think we want merge this into `mpldev` branch and not `main`.

0ctagon:
After reviewing the new svgs without modifying mplhep ([fa01534](https://github.com/cyrraz/plothist/pull/177/commits/fa015348825da20f06fd830c2b088576d182ecc4)), I see 3 main differences, and I fixed 2 of them in the last commit ([c0a124b](https://github.com/cyrraz/plothist/pull/177/commits/c0a124b9508df5bfd6ef6f68ee7c0eb9a37ac0d7)) by modifying a bit mplhep:

1. When stacked histograms are plotted, `ax.legend()` is not putting them in the right order in the legend (top to bottom). This was easily fixed by adding:

in `utils.py`

```
def stack(*plottables):
    plottables = plottables[::-1]
    ...
    plottables = plottables[::-1]
    return plottables
```

and in `plot.py`

```
if sort is not None:
   ...
elif stack:
    order = np.argsort(label)[::-1]
    plottables = [plottables[ix] for ix in order]
    _chunked_kwargs = [_chunked_kwargs[ix] for ix in order]
    _labels = [_labels[ix] for ix in order]
    if not "color" in kwargs:
        # Inverse default color cycle
        _colors = itertools.cycle(plt.rcParams["axes.prop_cycle"][len(plottables)-1::-1])
        for i in range(len(plottables)):
            _chunked_kwargs[i].update(next(_colors))
```

this modification mimics the matplotlib logic, and `ax.legend()` will behave the same.

2. By default, `step` style will display variances of the `Plotttable` if they exist. I don't think we want this to be the default. I propose that if `yerr` is `True`, then the variances are displayed. This was easily fixed by modifying the `do_errors` logic. A side effect of the `step` style is the representation in the legend. Before we had a square, now it's a line. Both seems fine to me.

3. The plots where the histograms are next to each others is not as straight forward, but it can be added with a custom style keyword like `step` or `fill`.

I also removed the `_kwargs = soft_update_kwargs(_kwargs, {"linewidth": 1.5})` to match the same one as before. Otherwise it agrees quite well :smiley: 

(The pytest not agreeing is because it is making the plot with the released version of matplotlib, not the one I modified locally. We could put those modifications on a mpldev branch in mplhep and install this branch in the gha)

0ctagon:
Replacing `plot_hist_uncertainties` was straight forward, I only needed to modify:
```
        err_defaults = {
            "linestyle": "none",
            "marker": ".",
            "markersize": 6.0, <------
            "elinewidth": 1.5, <------
        }
```
and no plot changed.

For `plot_error_hist`, I kept the uncertainty calculation method, which will have to be modified when we remove `boost-histogram` dependency, otherwise it was also quite straight forward:

```
        band_defaults = {
            "edgecolor": "dimgrey", <----
            "hatch": "////", <----
            "fill": False, <----
            "lw": 0, <----
        }
```

0ctagon:
[2a45b94](https://github.com/cyrraz/plothist/pull/177/commits/2a45b94b2cb2677b612eab80d3fd2bf3079c24d2) implemented `histtype=side` in `mplhep`. By default it is filled, but it also supports `step` and `step` with `yerr` representation.

0ctagon:
I updated the figures to correspond to the output of the current mplhep/plothist code. I think some default parameters could be changed, like the dot size or the grayness of the hashing. I would also change our default histograms with something without errors by default, so it doesn't overcrowd the plots. But for now every examples are working well with only using mplhep as a backend for plotting. I think we should merge this before opening another PR to focus on removing boost-histogram dependency (which will be easier to review with this PR merged already).

cyrraz:
@0ctagon If you are happy with this PR, could you please approve and merge this? I would like to start working on removing the boost_histogram dependency.

0ctagon:
@cyrraz ok let's merge, keeping in mind some code cleaning + some style fixing.

PR #179, opened 2024-12-12
feat: increase bh dependency limit

0ctagon:
We don't need to be that strict with bh version anymore.


0ctagon:
Also, [bh removed Python3.7 support in 1.5.0](https://github.com/scikit-hep/boost-histogram/releases/tag/v1.5.0). We could also remove it from your gha.

cyrraz (approved):

PR #180, opened 2025-02-07
fix version comparison when making examples, update precommit hooks, …

cyrraz:
* Fix version comparison when making examples
* Update pre-commit hooks
* Run pre-commit actions
* Update figures to matplotlib 3.10.0
* Require python>=3.9, test up to python3.13

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -3,37 +3,35 @@ name: Python package
 
 on:
   push:
-    branches: [ "main" ]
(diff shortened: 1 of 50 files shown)


0ctagon (approved):
We should also update the figure svg here https://github.com/cyrraz/plothist/pull/177

cyrraz:
> We should also update the figure svg here #177.
> 
> Do we publish a small release?

Yes, I will open a new PR for the release

PR #185, opened 2025-03-06
refactor: remove useless comments and reactivate checks

cyrraz:
* refactor: remove useless comments and reactivate checks

### DIFF
--- a/src/plothist/comparison.py
+++ b/src/plothist/comparison.py
@@ -95,13 +95,12 @@ def _check_binning_consistency(hist_list):
         If the histograms do not share the same dimensionality or if their bins are not equal.
 
     """
-    pass
-    # for h in hist_list:
-    #     if not len(h.axes) == len(hist_list[0].axes):
-    #         raise ValueError("Histograms must have same dimensionality.")
-    #     for i in range(len(h.axes)):
-    #         if not h.axes[i] == hist_list[0].axes[i]:
-    #             raise ValueError("The bins of the histograms must be equal.")
+    for h in hist_list:
+        if not len(h.axes) == len(hist_list[0].axes):
+            raise ValueError("Histograms must have same dimensionality.")
+        for i in range(len(h.axes)):
+            if not h.axes[i] == hist_list[0].axes[i]:
+                raise ValueError("The bins of the histograms must be equal.")
 
 
 def get_ratio_variances(h1, h2):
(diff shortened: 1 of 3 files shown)


src/plothist/histogramming.py
```diff
@@ -242,12 +242,11 @@ def _check_counting_histogram(hist):
         If the histogram is not a counting histogram.
 
     """
-    pass
-    # if hist.kind != bh.Kind.COUNT:
-    #     raise ValueError(
-    #         f"The histogram must be a counting histogram, but the input histogram has kind {hist.kind}."
-    #     )
-    # return
+    if hist.kind != bh.Kind.COUNT:
```
0ctagon, line 245: Need to see how / if we enforce this without bh

src/plothist/comparison.py
```diff
@@ -95,13 +95,12 @@ def _check_binning_consistency(hist_list):
...
 
     """
-    pass
-    # for h in hist_list:
-    #     if not len(h.axes) == len(hist_list[0].axes):
-    #         raise ValueError("Histograms must have same dimensionality.")
-    #     for i in range(len(h.axes)):
-    #         if not h.axes[i] == hist_list[0].axes[i]:
-    #             raise ValueError("The bins of the histograms must be equal.")
+    for h in hist_list:
```
0ctagon, line 98: Need to see how / if we enforce this without bh

0ctagon (approved):

src/plothist/histogramming.py
cyrraz: Yes, I want to keep it as long as possible, but we may remove it later.

src/plothist/comparison.py
cyrraz: Yes, I want to keep it as long as possible, but we may remove it later.

src/plothist/plotters.py
```diff
@@ -60,45 +61,32 @@ def create_comparison_figure(
...
+def plot_error_hist(
+    hist, ax, uncertainty_type="symmetrical", w2method="poisson", **kwargs
+):
+    _check_uncertainty_type(uncertainty_type)
+
+    if kwargs.get("density", False):
+        hist = hist.copy()
+        hist *= 1 / (hist.values() * hist.axes[0].widths).sum()
+
+    if uncertainty_type == "symmetrical":
```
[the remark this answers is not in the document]
cyrraz: Yes

PR #187, opened 2025-03-13
Remove boost histogram dependency 2

cyrraz:
This PR defines EnhancedNumPyPlottableHistogram and uses it instead of boost-histogram.

The goal is to replace the dependency on boost-histogram by a dependency on uhi.


src/plothist/histogramming.py
```diff
@@ -1,8 +1,74 @@
 import boost_histogram as bh
+from uhi.numpy_plottable import NumPyPlottableHistogram, Kind
 import numpy as np
 import warnings
 
 
+class EnhancedNumPyPlottableHistogram(NumPyPlottableHistogram):
```
0ctagon, line 7: Do we need some automatic tests on this class? Simple error checks or value/variance checks like we did in `test_histogramming.py`?

0ctagon (changes requested):
I don't see changes in `create_axis()`, could we also modify it in this PR? It's only used in tests and in the registry to get upper and lower edges.

src/plothist/histogramming.py
cyrraz: Done

cyrraz:
> Impressive work!!
> 
> I don't see changes in `create_axis()`, could we also modify it in this PR? It's only used in tests and in the registry to get upper and lower edges.

Done

tests/test_histogramming.py
```diff
@@ -45,9 +44,9 @@ def test_upper_edge_exclusive():
 
     data = [0] * 3 + [1] * 4 + [2] * 5 + [3] * 6 + [4] * 7 + [5] * 8
 
-    a1 = create_axis(5, (0, 5), overflow=True)
-    a2 = create_axis([0, 1, 2, 3, 4, 5], overflow=True)
-    a3 = create_axis(5, data=data, overflow=True)
+    a1 = bh.axis.Regular(5, 0, 5, overflow=True)
```
0ctagon, line 47: Is this test still required? We are not testing plothist here.

0ctagon (approved):
Minor comment, merge whenever!

tests/test_histogramming.py
cyrraz: True, but I would leave it there to check if there is any change in the bh implementation in the future. Since plothist and mplhep will want to be compatible with boost-histogram and Hist

PR #188, opened 2025-03-18
chore: switch to uv and hatchling

cyrraz:
* chore: switch to uv and hatchling
* doc: update installation doc


docs/usage/installation.rst
```diff
@@ -19,13 +19,13 @@ Using PyPI (stable version)
 Install ``plothist``
 --------------------
 
-1.  (Optional) If you use a local environment, activate it.
-
-    For example, run:
+2.  (Optional) Setup and activate a virtual environment:
```
0ctagon, line 22: This should be 1., or maybe 0. (and shift the rest)?

docs/usage/installation.rst
```diff
@@ -58,43 +58,49 @@ To update ``plothist`` to its latest stable version, follow the instructions abo
...
-       pip3 install flit
+2. Install ``uv`` following the instructions in the `uv documentation <https://docs.astral.sh/uv/getting-started/installation/>`_.
 
-3. Clone the plothist repository:
+3. Run the following command to install the package:
 
     .. code-block:: bash
 
-       git clone git@github.com:cyrraz/plothist.git
+       uv pip install --editable ".[dev,test]"
```
0ctagon, line 84: This is what I got running it for the first time:

```
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

If I add `--system` it gives (might be a me problem):
```
error: Failed to install: plothist-1.3.0-py3-none-any.whl (plothist==1.3.0 (from file:///home/plothist))
  Caused by: failed to create directory `/path/to/python3/dist-packages/plothist-1.3.0.dist-info`: Permission denied (os error 13)
```

Running `uv venv` before works, but gives these two warnings:
```
warning: The package `plothist @ file:///home/plothist` does not have an extra named `dev`
warning: The package `plothist @ file:///home/plothist` does not have an extra named `test`
```
So the `[dev,test]` might not be required?

docs/usage/installation.rst
```diff
@@ -58,43 +58,49 @@ To update ``plothist`` to its latest stable version, follow the instructions abo
...
+
+The previous command automatically creates a virtual environment ``.venv`` and installs the package in editable mode with the development and test dependencies.
```
0ctagon, line 86: We might want to add a small note that the user needs to activate the venv or use uv run

docs/usage/installation.rst
```diff
@@ -58,43 +58,49 @@ To update ``plothist`` to its latest stable version, follow the instructions abo
...
 
-   pip3 install --upgrade [--user] plothist
-
+    python3 -m pip install --upgrade [--user] plothist
 
-Using Flit (development version)
-================================
 
-1.  (Optional) If you use a local environment, activate it.
+Install the development version
```
0ctagon, line 64: We might want to specify that `pytest` is good to have with development version? It's not in our requirements.

docs/usage/installation.rst
0ctagon, line 84: `plothist_make_examples` doesn't seem to replace the images in `img/`.

docs/usage/installation.rst
cyrraz: It is installed when we do ``pip install --editable ".[dev,test]"``

docs/usage/installation.rst
cyrraz: Please try again from scratch with the updated instructions

cyrraz: Please try again from scratch with the updated instructions

docs/usage/installation.rst
cyrraz: Done

0ctagon (approved):

PR #199, opened 2025-03-27
style: add many linter rules and apply these rules

cyrraz:
Add many linter rules and apply these rules
```
[tool.ruff.lint]
extend-select = [
    "B",           # flake8-bugbear
    "I",           # isort
    "C4",          # flake8-comprehensions
    "ISC",         # flake8-implicit-str-concat
    "G",           # flake8-logging-format
    "PGH",         # pygrep-hooks
    "PIE",         # flake8-pie
    "RUF",         # Ruff-specific
    "UP",          # pyupgrade
    "YTT",         # flake8-2020
    "EXE",         # flake8-executable
    "NPY",         # NumPy specific rules
    "TID251",      # flake8-tidy-imports
    "SIM",         # flake8-simplify
    "RET",         # flake8-return
    "PD",          # pandas-vet
]
ignore = [
    "E402",  # module level import not at top of file
    "NPY002",  # Replace legacy `np.random` call with `np.random.Generator`
    "PD901",  # Avoid using the generic variable name `df` for DataFrames
]
```

In the future, we may work at removing the "ignore" and consider adding the following rules, but this will require some work.
```
"T20" # flake8-print
"PL" # pylint
"PT" # flake8-pytest-style
```


cyrraz:
Thank you @Copilot, I have removed the duplicate.

src/plothist/plotters.py
```diff
@@ -840,6 +847,16 @@ def plot_model(
...
+    if unstacked_kwargs_list is None:
+        unstacked_kwargs_list = []
+    if stacked_kwargs is None:
+        stacked_kwargs = {}
+    if unstacked_components is None:
+        unstacked_components = []
+    if stacked_components is None:
+        stacked_components = []
 
     # Create copies of the kwargs arguments passed as lists/dicts to avoid modifying them
```
0ctagon, line 861: Is this still useful?

src/plothist/plotters.py
```diff
@@ -1068,6 +1085,22 @@ def plot_data_model_comparison(
...
+    if unstacked_kwargs_list is None:
+        unstacked_kwargs_list = []
+    if stacked_kwargs is None:
+        stacked_kwargs = {}
+    if unstacked_components is None:
+        unstacked_components = []
+    if stacked_components is None:
+        stacked_components = []
+
+    # Create copies of the kwargs arguments passed as lists/dicts to avoid modifying them
```
0ctagon, line 1099: is this still useful?

src/plothist/scripts/install_latin_modern_fonts.py
```diff
@@ -115,11 +115,14 @@ def install_latin_modern_fonts():
                 attempt += 1
                 time.sleep(attempt)
                 subprocess.run(
-                    ["rm", "-f", (font_directory / f"latin-modern-{lm}.zip")]
+                    ["rm", "-f", (font_directory / f"latin-modern-{lm}.zip")],
+                    check=False,
```
0ctagon, line 119: Check is false by default, why this change? (same question for the other times this appears)

src/plothist/scripts/make_examples.py
```diff
@@ -3,6 +3,7 @@
 import subprocess
 import warnings
 
+import packaging.version as version
```
0ctagon, line 6: `from packaging import version` ?

src/plothist/scripts/make_examples.py
```diff
@@ -204,6 +205,9 @@ def make_examples(no_input=False, check_svg=False, print_code=False):
             fail(
                 f"The number of images has changed. Please run `plothist_make_examples`, check the new images and commit them if they are correct. New images:\n{set(new_img_hashes.keys()) - set(img_hashes.keys())}"
             )
+            return None
+        return None
+    return None
```
0ctagon, line 210: The rule is that each function should return something?

cyrraz: https://docs.astral.sh/ruff/rules/implicit-return/

src/plothist/plotters.py
cyrraz: Yes, if the original dictionaries are modified within the function, it could lead to unexpected behavior in other parts of the code that use the same dictionaries.

src/plothist/scripts/install_latin_modern_fonts.py
cyrraz: https://docs.astral.sh/ruff/rules/subprocess-run-without-check/

src/plothist/scripts/make_examples.py
cyrraz: Yes, will change this

PR #207, opened 2025-04-01
docs: remove trailing whitespace from the badges

0ctagon:
Remove the annoying underlined whitespaces from the badges, without altering the badges design.


cyrraz (approved):
Thank you

PR #215, opened 2025-06-17
style: add type hinting and fix errors raised by mypy

cyrraz:
This PR adds type hinting and fix errors raised by `mypy`.

### DIFF
--- a/src/plothist/plothist_style.py
+++ b/src/plothist/plothist_style.py
@@ -281,16 +281,14 @@ def add_text(
     }
 
     if isinstance(x, str):
-        resolved_x = x_values.get(x)
-        if resolved_x is None:
-            raise ValueError(f"{x} is not a valid x position")
-        x = resolved_x
+        if x not in x_values:
+            raise ValueError(f"{x!r} is not a valid x position.")
+        x = x_values[x]
 
     if isinstance(y, str):
-        resolved_y = y_values.get(y)
-        if resolved_y is None:
-            raise ValueError(f"{y} is not a valid y position")
-        y = resolved_y
+        if y not in y_values:
+            raise ValueError(f"{y!r} is not a valid y position.")
+        y = y_values[y]
 
     t = ax.text(
         x,
--- a/src/plothist/plothist_style.py
+++ b/src/plothist/plothist_style.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 
-def set_style(style="default"):
+def set_style(style: str = "default") -> None:
     """
     Set the plothist style.
 
@@ -39,15 +39,15 @@ def set_style(style="default"):
 
 
 def cubehelix_palette(
-    ncolors=7,
-    start=1.5,
-    rotation=1.5,
-    gamma=1.0,
-    hue=0.8,
-    lightest=0.8,
-    darkest=0.3,
-    reverse=True,
-):
+    ncolors: int = 7,
+    start: float = 1.5,
+    rotation: float = 1.5,
+    gamma: float = 1.0,
+    hue: float = 0.8,
+    lightest: float = 0.8,
+    darkest: float = 0.3,
+    reverse: bool = True,
+) -> list[tuple[float, float, float]]:
     """
     Make a sequential palette from the cubehelix system, in which the perceived brightness is linearly increasing.
     This code is adapted from seaborn, which implements equation (2) of reference [1] below.
@@ -75,9 +75,8 @@ def cubehelix_palette(
 
     Returns
     -------
-    list of RGB tuples
+    list[tuple[float, float, float]]
         The generated palette of colors represented as a list of RGB tuples.
-
 
     References
     ----------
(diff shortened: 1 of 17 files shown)


tests/test_efficiency.py
```diff
@@ -54,15 +54,15 @@ def test_efficiency_weighted_histograms():
     assert str(err.value) == error_msg
 
 
-def simple_efficiency_uncertainty(total, sample):
+def simple_efficiency_uncertainty(total, sample) -> None:
```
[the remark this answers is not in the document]
cyrraz: Thank you, cop, I have fixed this.

src/plothist/plothist_style.py
```diff
@@ -279,14 +281,16 @@ def add_text(
     }
 
     if isinstance(x, str):
-        x = x_values.get(x)
-        if x is None:
-            raise ValueError(f"{x} not a float or a valid position")
+        resolved_x = x_values.get(x)
+        if resolved_x is None:
+            raise ValueError(f"{x} is not a valid x position")
+        x = resolved_x
```
0ctagon, line 287: Which rules are imposing this change?

src/plothist/plotters.py
```diff
@@ -101,33 +100,33 @@ def plot_hist(hist, ax, **kwargs):
...
     Plot a 2D histogram using a pcolormesh plot and add a colorbar.
 
     Parameters
     ----------
     hist : boost_histogram.Histogram
         The 2D histogram to plot.
-    fig : matplotlib.figure.Figure, optional
-        The Figure instance for plotting. If fig, ax and ax_colorbar are None, a new figure will be created. Default is None.
-    ax : matplotlib.axes.Axes, optional
-        The Axes instance for plotting. If fig, ax and ax_colorbar are None, a new figure will be created. Default is None.
```
0ctagon, line 122: Maybe add a small sentence saying that `fig`, `ax` and `ax_colorbar` are either provided by the user or `None`, other cases will raise an error.

src/plothist/plotters.py
```diff
@@ -101,33 +100,33 @@ def plot_hist(hist, ax, **kwargs):
...
+        The Axes instance for plotting.
+    ax_colorbar : matplotlib.axes.Axes | None, optional
+        The Axes instance for the colorbar.
+    pcolormesh_kwargs : dict | None, optional
+        Additional keyword arguments forwarded to ax.pcolormesh().
+    colorbar_kwargs : dict | None, optional
+        Additional keyword arguments forwarded to ax.get_figure().colorbar().
     square_ax : bool, optional
-        Whether to make the main ax square (default is True).
+        Whether to make the main ax square. Default is True.
```
0ctagon, line 129: Why do we specify here that the `Default is True`, but everywhere else we removed the `Default is ...` sentence?

src/plothist/plotters.py
```diff
@@ -375,14 +386,14 @@ def plot_error_hist(hist, ax, uncertainty_type="symmetrical", density=False, **k
     ax : matplotlib.axes.Axes
         The Axes instance for plotting.
     uncertainty_type : str, optional
-        What kind of bin uncertainty to use for hist: "symmetrical" for the Poisson standard deviation derived from the variance stored in the histogram object, "asymmetrical" for asymmetrical uncertainties based on a Poisson confidence interval. Default is "symmetrical".
+        Type of bin uncertainty to use: "symmetrical" or "asymmetrical".
```
0ctagon, line 389: Less information for the user is usually not the best. Is there any reasons for this removal?

src/plothist/plotters.py
```diff
@@ -459,20 +470,20 @@ def plot_two_hist_comparison(
...
+    ylabel : str | None, optional
+        The label for the y-axis.
     h1_label : str, optional
         The label for the first histogram. Default is "h1".
     h2_label : str, optional
         The label for the second histogram. Default is "h2".
-    fig : matplotlib.figure.Figure or None, optional
-        The figure to use for the plot. If fig, ax_main and ax_comparison are None, a new figure will be created. Default is None.
-    ax_main : matplotlib.axes.Axes or None, optional
-        The main axes for the histogram comparison. If fig, ax_main and ax_comparison are None, a new axes will be created. Default is None.
```
0ctagon, line 473: Maybe add a small sentence saying that `fig`, `ax` and `ax_colorbar` are either provided by the user or `None`, other cases will raise an error.

src/plothist/variable_registry.py
```diff
@@ -76,7 +81,7 @@ def create_variable_registry(
         variable name in data.
     bins : int
         Number of bins, default is 50.
-    range: list of two float
+    range: list[float]
```
0ctagon, line 84: Is there a way to say `len(list) = 2`? (the text before was giving this info)

0ctagon (commented):
Nice addition.

src/plothist/plothist_style.py
0ctagon:
```
if isinstance(x, str):
    if x not in x_values:
        raise ValueError(f"{x!r} is not a valid x position")
    x = x_values[x]   
```

I think this solution is a bit nicer since it doesn't introduce a new temporary variable.

cyrraz:
> Before merging, check the 2 comments of copilot (the misspell of `get_efficency` is a funny overlook) and my suggestion on the mypy error.

Thank you, I have implemented the comments and will merge as soon as the tests are done.

PR #221, opened 2025-07-10
refactor: remove make-example script and svg (exception: README) and update uv.lock

cyrraz:
This PR removes make-example script and svg images (exception: the svg in the README) and update uv.lock


uv.lock
0ctagon, line 1: The modifications here are still a mystery here.

0ctagon (approved):
Clean.
How do we update the figures from the README? they need to fetch the metadata from `plothist_utils`.

0ctagon:
Something like [this](https://github.com/scikit-hep/mplhep/blob/main/.github/workflows/ci.yml#L60) would be nice to debug errors.

PR #223, opened 2025-07-24
fix: improve asymmetrical uncertainty calculations

0ctagon:
Asymmetrical uncertainties were not behaving correctly when:
- either the upper or lower uncertainty was `nan`, the other one was not displayed (for points a 0, see attached screenshots)
- For points at 0, we should take one-sided upper limit (upper uncertainty value from 1.83 to 1.15).

Screenshot from 2025-07-24 17-01-48
Screenshot from 2025-07-24 17-01-33


src/plothist/comparison.py
```diff
@@ -77,11 +77,27 @@ def get_asymmetrical_uncertainties(
         raise ValueError(
             "Asymmetrical uncertainties can only be computed for an unweighted histogram."
         )
+
     conf_level = 0.682689492
     alpha = 1.0 - conf_level
-    n = hist.values()
-    uncertainties_low = n - stats.gamma.ppf(alpha / 2, n, scale=1)
-    uncertainties_high = stats.gamma.ppf(1 - alpha / 2, n + 1, scale=1) - n
+    n = np.asarray(hist.values())
```
cyrraz, line 83: I think `hist.values()` is already returning a `np.array`, no?

src/plothist/comparison.py
```diff
@@ -77,11 +77,27 @@ def get_asymmetrical_uncertainties(
...
+
+    lower = np.zeros_like(n, dtype=float)
+    upper = np.zeros_like(n, dtype=float)
+
+    # One-sided upper limit for n == 0
+    upper[n == 0] = stats.gamma.ppf(conf_level, 1)
```
cyrraz, line 89: I think we should not derive from what ROOT does. ROOT did have the same hesitation as we have now, but decided to have no particular behavior for the case n=0:

https://root.cern.ch/doc/master/TH1_8cxx_source.html#l09158

cyrraz (changes requested):
Thank you for the PR, I agree that we should solve the but coming when we have nan values, butI do not think we should have a particular behavior when n=0. See my comment.

cyrraz:
> You can find attached the different behavior from ROOT, Roofit and mplhep. We can see that mplhep (and Hist) and TH1 don't have a special case for 0, while roofit and this PR has (for roofit I think it's defined [here](https://root.cern.ch/doc/master/RooDataHist_8cxx_source.html#l01556) -> [here](https://root.cern.ch/doc/master/RooHistError_8cxx_source.html) + [here](https://root.cern.ch/doc/master/RooHistError_8h_source.html) but it's not the easiest to understand). 
> 
> 
> 
> I do like the logic behind the roofit error: the frequentist approach says you have 68% coverage between the error values. For 0 since you don't have the lower bound, you don't have 68% but 84%, hence the one sided definition. We can also ask mplhep why they went with this definition and not the roofit one.
> 
> 
> 
> Screenshot from 2025-07-25 12-02-59
> 
> Screenshot from 2025-07-25 12-17-10
> 
> Screenshot from 2025-07-25 12-17-16
> 
> 

Could you please attach plots in colors?

cyrraz (approved):
I pushed a new commit and I approve this PR

0ctagon:
I added an argument for the users to choose between how they treat the uncertainty for empty bins. Since this is an expert level argument, doc has not been updated for `plot_error_hist()` but the statistical section is more explained.

src/plothist/comparison.py
```diff
@@ -42,11 +49,13 @@ def _is_unweighted(hist: bh.Histogram) -> bool:
     bool
         True if the histogram is unweighted, False otherwise.
     """
-    return np.allclose(hist.variances(), hist.values())
+    print(hist.variances(), hist.values())
```
cyrraz: Remove print

tests/test_comparison.py
```diff
@@ -4,18 +4,28 @@
...
+    get_asymmetrical_uncertainties,
     get_comparison,
 )
 
 
 def test_check_uncertainty_type_invalid():
     with pytest.raises(
         ValueError,
-        match="Uncertainty type invalid_type not valid. Must be 'symmetrical' or 'asymmetrical'.",
+        match="Uncertainty type invalid_type not valid. Must be in \['symmetrical', 'asymmetrical', 'asymmetrical_double_sided', 'asymmetrical_one_sided'\].",
```
cyrraz, line 15: Fix warnings.

src/plothist/comparison.py
```diff
@@ -189,7 +217,9 @@ def get_pull(
     _check_counting_histogram(h2)
 
     if h1_uncertainty_type == "asymmetrical":
-        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(h1)
+        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(
+            h1, h1_uncertainty_type
```
cyrraz, line 220: This is a bug. If the user pass `asymmetrical_double_sided`, the uncertainty will be treated as `symmetrical` because of the `if h1_uncertainty_type == "asymmetrical":`

src/plothist/comparison.py
```diff
@@ -247,7 +277,9 @@ def get_difference(
     difference_values = h1.values() - h2.values()
 
     if h1_uncertainty_type == "asymmetrical":
-        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(h1)
+        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(
+            h1, h1_uncertainty_type
```
cyrraz, line 280: This is a bug. If the user pass `asymmetrical_double_sided`, the uncertainty will be treated as `symmetrical` because of the `if h1_uncertainty_type == "asymmetrical":`

src/plothist/comparison.py
```diff
@@ -405,7 +437,9 @@ def get_ratio(
     ratio_values = np.where(h2.values() != 0, h1.values() / h2.values(), np.nan)
 
     if h1_uncertainty_type == "asymmetrical":
-        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(h1)
+        uncertainties_low, uncertainties_high = get_asymmetrical_uncertainties(
```
cyrraz, line 439: This is a bug. If the user pass `asymmetrical_double_sided`, the uncertainty will be treated as `symmetrical` because of the `if h1_uncertainty_type == "asymmetrical":`

src/plothist/comparison.py
```diff
@@ -72,23 +83,40 @@ def get_asymmetrical_uncertainties(
...
+            "Invalid uncertainty type 'symmetrical' for asymmetrical uncertainties."
+        )
+
     alpha = 1.0 - 0.682689492
     n = hist.values()
 
     lower_bound = np.zeros_like(n, dtype=float)
     upper_bound = np.zeros_like(n, dtype=float)
 
+    tail_probability = alpha / 2
```
cyrraz, line 103: define it just below alpha

src/plothist/comparison.py
```diff
@@ -72,23 +83,40 @@ def get_asymmetrical_uncertainties(
...
-    lower_bound[n > 0] = stats.gamma.ppf(q=alpha / 2, a=n[n > 0], scale=1)
-    upper_bound[n > 0] = stats.gamma.ppf(q=1 - alpha / 2, a=n[n > 0] + 1, scale=1)
-    # One-sided upper limit for n == 0
-    upper_bound[n == 0] = stats.gamma.ppf(q=1 - alpha, a=1, scale=1)
+    lower_bound[n > 0] = stats.gamma.ppf(q=tail_probability, a=n[n > 0], scale=1)
+    upper_bound[n > 0] = stats.gamma.ppf(
+        q=1 - tail_probability, a=n[n > 0] + 1, scale=1
+    )
+
+    if uncertainty_type == "asymmetrical_double_sided":
```
cyrraz, line 111: I find confusing to change the tail probability in this if else.

Please do something like

```
if uncertainty_type == "asymmetrical_double_sided":
    # Two-sided Garwood intervals for n == 0
    upper_bound[n == 0] = stats.gamma.ppf(q=1 - tail_probability, a=1, scale=1)
else:
    # One-sided upper limit for n == 0
    upper_bound[n == 0] = stats.gamma.ppf(q=1 - 2*tail_probability, a=1, scale=1)
```

I think this helps to see that we are doing something only for `n==0`.

cyrraz (changes requested):
Thank you for the update, I found a bug in the latest commits.

tests/test_comparison.py
0ctagon: What is the problem?

cyrraz, line 15: Easier:

`match="Uncertainty type invalid_type not valid. Must be in"`

No need to hard code the end of the string in the match.

cyrraz: See my other comment, where I propose another way to fix this.

tests/test_comparison.py
```diff
@@ -4,18 +4,28 @@
...
     ):
         _check_uncertainty_type("invalid_type")
 
 
+def test_get_asymmetrical_uncertainties_invalid():
```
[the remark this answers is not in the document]
0ctagon: No need, the `_check_uncertainty_type` already catches everything except `symmetrical` that we test here.

src/plothist/examples/utility/uncertainty_types.py
```diff
@@ -0,0 +1,120 @@
...
+
+    for i, idx in enumerate(indices):
+        if i >= len(values):
+            break
+        arr[idx] = values[i]
+
+    return arr
+
+
+def make_grouped_edges(
```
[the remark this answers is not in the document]
0ctagon: Unfortunately `plot_error_hist` can only take 1 hist and uses `ax.errorbar` where you specify by hand `x, y`; so no tricks like `ax.hist` unless we change `plot_error_hist`.

PR #260, opened 2026-05-14
chore: update uv.lock

0ctagon:
Fixing vulnerabilities.


cyrraz (approved):
Thank you for fixing vulnerabilities.

PR #264, opened 2026-06-12
fix: prevent input mutation and figure leaks in plotters

0ctagon:
:robot: Claude Opus

- plot_hist: fix np.nan_to_num(values, 0) passing 0 as copy arg (in-place mutation)
- plot_model / plot_data_model_comparison: deep-copy unstacked_kwargs_list dicts
- plot_model: use list comprehension instead of [{}]*n
- plot_model: compute sum(components) once
- plot_data_model_comparison: close dummy figure in plot_only path
- create_comparison_figure: squeeze=False+ravel(), default height_ratios from nrows
- plot_function: rename shadowed loop variable func -> f

### DIFF
--- a/tests/test_plotters.py
+++ b/tests/test_plotters.py
@@ -289,3 +289,71 @@ def test_plot_data_model_comparison_cases() -> None:
         plot_data_model_comparison(
             data_hist=h_1d, stacked_components=[func], plot_only="invalid"
         )
+
+
+def test_plot_hist_does_not_mutate_nan_histogram() -> None:
+    """
+    Test that plot_hist does not modify the input histogram's storage when it
+    contains NaN bin values (the NaN-to-zero replacement must operate on a copy).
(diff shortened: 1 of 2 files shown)


tests/test_plotters.py
```diff
@@ -289,3 +289,71 @@ def func(x):
...
             data_hist=h_1d, stacked_components=[func], plot_only="invalid"
         )
+
+
+def test_plot_hist_does_not_mutate_nan_histogram() -> None:
+    """
+    Test that plot_hist does not modify the input histogram's storage when it
+    contains NaN bin values (the NaN-to-zero replacement must operate on a copy).
+    """
+    from plothist import plot_hist
```
cyrraz: imports should be at the beginning of the file

cyrraz (approved):
Thank you, please check a minor comment

